PR #4, opened 2020-01-07
Load macros with context, move path logic into macro

fofr:
- Make `page` available on the image macro
- Remove the need to reuse the same path manipulation logic in every post
- Simplify syntax in posts


paulrobertlloyd:
Doh! Is `context` an Eleventy variable, or a special Nunjucks one. Don’t think I’ve come across it before.

fofr:
`with context` is a nunjucks thing:
https://github.com/mozilla/nunjucks/pull/1209/files

PR #9, opened 2020-01-08
Simplify article template for GOV.UK parity, other small fixes

fofr:
Keep page templates inline with GOV.UK content, for example: https://www.gov.uk/vehicle-log-book

- Move date to bottom of the page
- Fix alignment of prose with related links
- Remove description – a search description is a summary serving a different purpose to a tagline

A lot of pages also open with content similar to the search description, meaning pages were unnecessarily repeating content. This avoids needing to rewrite them.

Couple of fixes @adamsilver requested:
- Rename product to "Becoming a teacher design history"
- Never link to current page in breadcrumbs

### DIFF
--- a/app/_components/article/_article.scss
+++ b/app/_components/article/_article.scss
@@ -1,28 +1,4 @@
-.app-article {
-  .app-prose {
-    @include govuk-responsive-margin(6, "top");
-  }
-}
-
-.app-article__header {
-  @include govuk-responsive-padding(2, "bottom");
-}
-
-.app-article__title {
-  @extend %govuk-heading-xl;
-  margin: 0 0 govuk-spacing(2);
-}
-
-.app-article__description {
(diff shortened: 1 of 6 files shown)


adamsilver:
Why is the date at the bottom?

fofr:
Design history has until now not had dates on pages. There wasn’t a strong need to add them on the content page. Dates were most useful when finding content. So I’m giving the content priority over the date, for now.

It's not clear what the date is. Is it published date or a rough date the design was worked on? It's usually been the latter.

See also issues like these:
Screen Shot 2020-01-08 at 14 48 44

paulrobertlloyd:
LGTM. One little issue with the longer product name is we should adjust the header breakpoints to prevent this from happening:

Screenshot 2020-01-08 at 15 09 43

fofr:
Will fix the header wrapping in a follow-up

PR #11, opened 2020-01-10
Add pagination component and paginate collections

paulrobertlloyd:
* Adds pagination component
* Adds pagination to collection pages (50 items/page). This means it currently only appears on the ‘Publish’ collection.

Styled using an amalgamation of pagination used on GOV.UK blogs (example: https://aphascience.blog.gov.uk/page/2/) and GOV.UK Whitehall (?) (example: https://www.gov.uk/youth-crime-prevention-programmes/how-young-people-are-put-on-a-programme).

Example:
Screenshot 2020-01-10 at 15 39 36

Example when no previous page:
Screenshot 2020-01-10 at 15 38 54

On my personal site, I rewrite the pagination URLs to use a query, i.e. from `/foo/bar/page2.html` to the more correct `/foo/bar?page=2`, but that can come later.

### DIFF
--- a/app/_components/_all.scss
+++ b/app/_components/_all.scss
@@ -9,6 +9,7 @@
 @import "gallery/gallery";
 @import "header/header";
 @import "masthead/masthead";
+@import "pagination/pagination";
 @import "prose/prose";
 @import "related/related";
 @import "screenshots/screenshots";
--- a/app/_components/pagination/_pagination.scss
+++ b/app/_components/pagination/_pagination.scss
@@ -1,0 +1,59 @@
+.app-pagination {
+  @include govuk-font($size: 19);
+  display: flex;
+  justify-content: space-between;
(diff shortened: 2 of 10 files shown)


adamsilver (commented):
Nice one Paul. Pagination sorely needed.

As this is for the design history I'd say this is okay but I worry about how it works when there's no previous/the fact that the pages and back/next links are centered/so far apart.

I'd suggest using this pagination control instead:
https://moj-design-system.herokuapp.com/components/pagination

It doesn't use SVG, everything is aligned left and there's an optional ‘showing’ note that goes on the right but we probs don't need that.

paulrobertlloyd:
@adamsilver Just the sort of helpful feedback I was expecting and hoping for. On it!

paulrobertlloyd:
Now uses the MoJ component wholesale. Not showing results counts (as not sure if 11ty provides them, certainly not easily).

PR #12, opened 2020-01-10
Update article footer appearance, add support for modified dates

paulrobertlloyd:
* Add support for `modified` front matter field for posts. If used, ‘Last updated’ will appear in post footer.
* Updated design of footer to match that on GOV.UK:

With published date:
Screenshot 2020-01-10 at 17 06 11

With published and last updated dates:
Screenshot 2020-01-10 at 17 05 50

### DIFF
--- a/app/_components/article/_article.scss
+++ b/app/_components/article/_article.scss
@@ -1,4 +1,14 @@
+.app-article__body {
+  @include govuk-responsive-margin(8, "bottom");
+
+  *:last-child {
+    margin-bottom: 0;
+  }
+}
+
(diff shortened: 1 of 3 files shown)


paulrobertlloyd:
@fofr Suggested using something like ‘Design proposed’ or ‘Design created’ instead of ‘Published’, which is a good point (it’s likely the dates we will want to use are those when a design was first proposed, not when history was published to the site). However, not all articles will necessarily be design related (performance posts, for example). Keeping the generic ‘Published’ for now; we can always update in a separate PR.

PR #15, opened 2020-01-10
Integrate Netlify CMS

paulrobertlloyd:
Access CMS at `/admin`. Use GitHub account to log in.

### List of posts:

collection

### Editing a post:

edit-post

## Current limitations

* Any macros used aren’t rendered in the preview (parsing Nunjucks macros in Markdown via the React components used by Netlify CMS gets icky, fast!)
* Markdown image markup (ie `image`) isn’t rendered in preview
* May also need oAuth permissions for NetlifyCMS on repo, not sure if they’ve been granted yet.

There’s [a lot that can be configured](https://www.netlifycms.org/docs/configuration-options/), including the preview and adding custom widgets, but probably best to start with something relatively basic and see if people find it useful, before doing anything too custom.


adamsilver (approved):

fofr:
This isn't working on Heroku because we can't get an auth provider set up. I'd like to revert this until we can get the feature running properly.

PR #27, opened 2020-01-21
Remove tags field from generator scripts

paulrobertlloyd:
Removes `tags:` and line break before content from generator scripts. For consistency, innit.


fofr:
The line break is where people should start typing. I wonder if that should be more explicit.

fofr (approved):

PR #32, opened 2020-01-22
User need macro

adamsilver:
Before:

image

After:

image

Used the style from the service manual

### DIFF
--- a/app/_components/_all.scss
+++ b/app/_components/_all.scss
@@ -16,3 +16,4 @@
 @import "screenshots/screenshots";
 @import "site-search/site-search";
 @import "summary-card/summary-card";
+@import "user-need/user-need";
--- a/app/_components/user-need/_user-need.scss
+++ b/app/_components/user-need/_user-need.scss
@@ -1,0 +1,9 @@
+.app-user-need {
+  background-color: govuk-colour('light-grey');
+  padding: govuk-spacing(5);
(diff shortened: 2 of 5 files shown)


fofr:
I think that needing to write the following:

```
{% from "user-need/macro.njk" import appUserNeed with context %}

{{appUserNeed({
  user: 'referee',
  goal: 'to provide a useful reference',
  outcome: 'I can help the candidate with their application and get back to my day as quickly as possible'
})}}
```

Instead of simply:

```
As a referee
I need to provide a useful reference
So that I can help the candidate with their application and get back to my day as quickly as possible
```

Is not the answer.

adamsilver:
I'd love to write that to achieve the same outcome. How do we do that?

paulrobertlloyd:
There’s always:

```html
<div class="app-user-need">
As a referee
I need to provide a useful reference
So that I can help the candidate with their application and get back to my day as quickly as possible
</div>
```

Feels icky, but works just as well.

fofr:
Custom markdown might look like:

```
{* need *}
As a referee
I need to provide a useful reference
So that I can help the candidate with their application and get back to my day as quickly as possible
{* endneed *}
```

adamsilver:
Put this to a vote on Slack and got the following results:

0/10 prefer option 1
7/10 prefer option 2
3/10 prefer option 3 which is a grey background and less whitespace between the paragraphs

PR #33, opened 2020-01-22
Hypothesis macro

adamsilver:
Before:

image

After:

image

I don't especially like the border split but throwing it in here early for a crit.

### DIFF
--- a/app/_components/_all.scss
+++ b/app/_components/_all.scss
@@ -8,6 +8,7 @@
 @import "figure/figure";
 @import "gallery/gallery";
 @import "header/header";
+@import "hypothesis/hypothesis";
 @import "masthead/masthead";
 @import "pagination/pagination";
 @import "prose/prose";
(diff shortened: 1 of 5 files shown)


fofr:
```
{{ appHypothesis({
  title: 'Actively refusing to give a reference',
  context: { text: 'Some referees may not want to, or be able to give a reference. This slows down the candidate\'s application process.' },
  if: 'we let users tell us that they won\'t give a reference',
  then: 'we\'ll be able to notify the candidate and get another referee quickly',
  measurements: [
    'referees click the refuse link in the email and confirm their refusal'
  ]
})}}
```

This isn't very user friendly.

paulrobertlloyd:
User-friendliness is the only downside of these macros. But! I wonder if we could create some corresponding custom widgets, and use those in the CMS? https://www.netlifycms.org/docs/custom-widgets/

fofr:
Rigidness is another downside.

PR #37, opened 2020-01-22
Add date field to posts, ensure future generated posts have dates too

paulrobertlloyd:
11ty can derive a post’s publish date from its filename, however Netlify CMS cannot. It looks for a `date` field in the frontmatter. Without that, it assumes todays date, and if making changes, will add that date to a post. So… added a `date` field to all previous posts. Which was fun.

Also updated the generator scripts to include this field also.

### DIFF
--- a/app/posts/apply-for-teacher-training/2019-02-19-apply-august-2018.md
+++ b/app/posts/apply-for-teacher-training/2019-02-19-apply-august-2018.md
@@ -1,6 +1,7 @@
 ---
 title: Apply – August 2018
 description: The Apply designs as they stood at the end of August 2018.
+date: 2019-02-19
 related:
   items:
   - text: Prototype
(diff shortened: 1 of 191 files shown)


app/posts/apply-for-teacher-training/2019-02-19-apply-august-2018.md
```diff
@@ -1,6 +1,7 @@
 ---
 title: Apply – August 2018
 description: The Apply designs as they stood at the end of August 2018.
+date: 2019-02-19
```
fofr: This date should probably be `2018-08-01`

fofr (approved):

app/posts/apply-for-teacher-training/2019-02-19-apply-august-2018.md
paulrobertlloyd: Updated all early Apply posts with dates that reflect titles.

PR #45, opened 2020-01-30
Allocations summary

jesseyuen:
This is a work in progress entry for the last Allocations summary. I'm waiting on the communications we sent to providers.


fofr (approved):
Updated date, also tweaked title of an earlier allocations post

PR #59, opened 2020-02-11
Use forked version of webshot

paulrobertlloyd:
I always seem to get the following error when trying to generate screenshots with the provided script:

```
const { Math, Object } = primordials;
                         ^
ReferenceError: primordials is not defined
```

Using a forked version of the `webshot` package (which has a newer version of `graceful-fs` as a dependency) fixes this issue.


fofr:
Ah, nice fix

PR #61, opened 2020-02-12
Remove express

adamsilver:
Remove Express

Change build commands to work without Express both locally and in live (Netlify).

Simplify commands

`npm start` now just watches the files
`npm run watch` is for when working on the JS and CSS which is rare.


paulrobertlloyd:
Before we merge this, think we should at least reinstate the redirects currently in `app.js` in a Netlify config file.

I’d be tempted to not change the image paths either, and instead use rewrites for now (I *think* that’s possible). We can then remove those rewrites once we’ve reinstated image resizing using a Netlify function (which I think we should do fairly soon, as the raw images are pretty big).

fofr:
@paulrobertlloyd This should do the trick: https://github.com/DFE-Digital/bat-design-history/pull/62

fofr:
Any reason this branch doesn't work on Heroku? The slug size is still below 500mb.

adamsilver:
@fofr I think because Heroku is (currently) set up to work with Express etc (buildpacks etc )

adamsilver:
I think we're good to go. Will leave the image rewriting/resizing for another PR.

PR #76, opened 2020-02-25
Provider needs

adamsilver:
This is a draft because it needs work to add genuine user needs to the codebase.

But this has provided a few mechanisms in order to do it.

I'm looking for a review on the implementation and conventions behind the mechanism.

Another thing: should I refer to needs as just need or needs as opposed to user needs and user need?

https://www.loom.com/share/75b6c67342df40e6b85f41db3468a18f


paulrobertlloyd:
Just to recap the feedback I gave in our call:
- Consistent use of `user-need` over `need` for the various parts of this feature.
- We could probably include the for loop for user needs in the/a macros, as we’d end up repeating that code several times.
- Use the document list component for listing entries relating to a particular need
- Use the user need code as the title
- We should also document the naming convention we’re using for needs.

Also to think about, but maybe in separate PRs:
- Look at reducing the overall size of user needs cards; is there a way we can reduce the amount of space they take up
- Think about how we might link to the user needs page. A link in the ‘Related links’ sidebar seems wrong. Maybe we need secondary navigation inside each service page?

adamsilver:
@paulrobertlloyd I've made those changes. Only thing is in the front matter of a need, the key is userneed` (not `user-need`) because that's invalid Yaml/front matter.

adamsilver:
@paulrobertlloyd I had to make the front matter key `userneed` (not `user-need`) because Nunjucks will throw when trying to read a variable with a dash in it.

scripts/screenshot.js
```diff
@@ -6,7 +6,7 @@
   Run:
   node scripts/screenshot.js name-of-directory
   or
-  node scripts/screenshot.js apply-for-teacher-training/name-of-directory
+  node scripts/screenshot.js manage-teacher-training-applications/improving-check-and-confirm-content
```
paulrobertlloyd: We shouldn’t track changes in this file (though we can’t ignore it, either).

app/_layouts/user-need.njk
```diff
@@ -0,0 +1,46 @@
...
+          description: userneed,
+          status: status
+        })}}
+
+        {{ appProse({
+          prose: content
+        }) }}
+
+        {% if collections[page.fileSlug].length %}
+          <h2 class="govuk-heading-l">Posts</h2>
```
paulrobertlloyd: The size of this heading should match that on the collections index layout:

```suggestion
          <h2 class="govuk-heading-l govuk-!-font-size-27">Posts</h2>
```

paulrobertlloyd (commented):
Just need some guidance about naming conventions for needs. Take it AN = Apply Need, MN = Manage Need? Why not candidate and provider again?

app/posts/apply-for-teacher-training/2020-01-21-suitability-to-work-with-children.md
```diff
@@ -41,11 +45,8 @@ Any 'disclosure' would be superseded by the DBS so it is rather redundant.
 
 ## User needs
 
-As a provider I need to know if a candidate has a history that would affect their application, so that I can either discuss and assist them in their journey, or reject them, if appropriate.
-
-As a candidate with a checkered history I need an opportunity to be honest about my past, so that I have a better chance of becoming a teacher.
-
-As a candidate with a checkered history I need to know what a DBS check is, and what it will say about me, so that I can make a decision about what to disclose beforehand.
+{% from "user-needs/macro.njk" import appUserNeeds %}
+{{ appUserNeeds({ items: collections['user-need'] | slugs(tags)}) }}
```
fofr, line 50: I think this can be simplified further to:
```
{% from "user-needs/macro.njk" import appUserNeeds with context %}
{{ appUserNeeds }}
```

Then within the user needs macro you will have the context of the current post, so you can hide `items: collections['user-need'] | slugs(tags)` away in the macro too.

The screenshots macro is an example of this working.

adamsilver:
@paulrobertlloyd if I made it CN for candidate need then there will be a mix up with candidate needs on Find and Apply.

PR #92, opened 2020-03-13
The summary card component

paulrobertlloyd:
Document the design of the summary card component, when and how to use it, and the different variants of it that are supported.


fofr:
Why has Netlify not generated a preview for this?

paulrobertlloyd:
@fofr Odd, it has but not updated GitHub. Bit flakey, it seems… Anyway: https://deploy-preview-92--bat-design-history.netlify.com

fofr (approved):

PR #102, opened 2020-03-25
Remove attempted Netlify CMS integration

paulrobertlloyd:
Removing files and changes required to integrate with Netlify CMS. Not only did it require a lot of changes, we could never get the CMS to properly authenticate with GitHub.

Also, and most importantly, we seem to be doing pretty okay with contributions from the wider team – although we should continue keeping an eye on this, and look to other ways we can improve and broaden access.


fofr (approved):
😞

PR #103, opened 2020-03-25
Add Markdown linter and lint markdown

paulrobertlloyd:
Linter can be run directly using `npm run lint:markdown` or as part of `npm run lint`.


fofr:
Nice :100:

What were all the `<!-- markdownlint-disable MD024 -->` for?

paulrobertlloyd:
The linter doesn’t see Nunjucks, so thinks there are multiple headings with the same name and at the same level on the page. Oh, and in one case, it warns about needing a space before a list, again, because it’s not understanding the template syntax. Did try rewriting the captions, then thought best to just add the exceptions.

fofr (approved):

PR #109, opened 2020-03-30
Sub navigation for application (Manage applications)

adamsilver:


paulrobertlloyd:
For ‘Concise header’, isn’t another one of the reasons for the change because the standard header has poor placement, visibility and colour contrast for the sub navigation links?

PR #128, opened 2020-04-29
Document tweaks to content for accepting or declining offers

fofr:
https://trello.com/c/FNV1k2V8/1237-apply-2-update-guidance-on-accepting-or-rejecting-offers

“Are you sure you want to accept offer” contained guidance on withdrawing other choices that’s not relevant when applying again (there’s only 1 choice).

“Are you sure you want to decline offer” contains guidance about applying again through UCAS, when this feature is launched apply again will not be specific to UCAS.

The updated content removes mention of:

- UCAS when suggesting a candidate can apply again after declining
- the caveat about “all other offers” when declining an offer on an application with 1 choice
- automatically withdrawing other choices when accepting an offer on an application with 1 choice

These changes should also apply to first applications that have only 1 choice.


paulrobertlloyd:
Not sure if it’s worth mentioning/addressing with this issue, but when I added the survey to withdrawal reason, after chatting with @EmmaFrith, we decided to remove (rather than edit) the similar guidance that appeared on that page. Not sure if that’s how we want to keep it? https://bat-design-history.netlify.app/apply-for-teacher-training/reason-for-withdraw#withdraw

paulrobertlloyd:
Withdrawing page previously said:

> If you withdraw this course choice (and all your other course choices), you can still apply for more courses this year through UCAS Teacher Training.

fofr:
Guidance also added to withdraw screens

PR #133, opened 2020-05-06
Manage notifications v1 added

jesseyuen:
This is a initial stub to introduce the early work done on managing notifications. I'll dig up and include the research findings and user feedback in the next edit - promise.


fofr (approved):
:+1:

PR #138, opened 2020-05-14
Use Eleventy Navigation to power breadcrumbs and site navigation

paulrobertlloyd:
* Replaces bespoke, error prone and unwieldy breadcrumb implementation with one powered by `eleventy-navigation`, [a first-party Eleventy plugin](https://www.11ty.dev/docs/plugins/navigation/).
* Given this plugin generates a map of all pages on the site, use this to power both the top-level section listing on the homepage, and list all pages on the sitemap
* Sum total of this change means less information is needed in frontmatter data to manage breadcrumbs. ~~That said, nested folders have proven somewhat problematic; might need a second crack at these scenarios to work out how deep hierarchies can be linked up correctly without having to override `eleventyNavigation` on sub-folder indices and directory data files.~~

Also…

* ~~Tidies up and simplifies some of the templating around user needs, and ensures they use pretty URLs~~ Moved to #139

(@fofr Wanted to test using this plugin on our own site given its complexity. Idea however is to add this into the boilerplate design history once all the wrinkles have been ironed out - breadcrumbs right now need too much manual intervention for them to work; hopefully using this plugin can fix that!)


paulrobertlloyd:
Realise these changes broke the search results, but in trying to fix that, broke everything else. 🤦🏼‍♂️ Basically, this is still causing me grief:

> nested folders have proven somewhat problematic

Combination of navigation plugin needing items to be excluded from a collection, but we want to hide sub folder items from collection listings. I need a pair of fresh eyes for this.

paulrobertlloyd:
Right, got this all working properly. A consequence of using this plugin however is that pages showing in breadcrumbs can’t be excluded from collections. But! Using `eleventyExcludeFromCollections` conflicted with this requirement, but there were other - arguably more suitable means – of achieving the same result.

Some rejigging of folders and directory data fixed this.

app/_components/breadcrumbs/template.njk
```diff
@@ -0,0 +1,11 @@
+{% set items = params.items | eleventyNavigationBreadcrumb(params.key) %}
+<div class="govuk-breadcrumbs{%- if params.classes %} {{ params.classes }}{% endif %}">
+  <ol class="govuk-breadcrumbs__list">
+  {%- for item in items %}
+    <li class="govuk-breadcrumbs__list-item">
+      <a class="govuk-breadcrumbs__link" href="{{ item.url | url | pretty }}">{{ item.title }}</a>
```
fofr, line 6: What's the `| pretty` doing here?

app/posts/find-teacher-training/2017-12-01-alpha.md
```diff
@@ -13,12 +11,16 @@ related:
       Password: beta
   - text: "Prototype code on Github"
     href: https://github.com/DFE-Digital/search-and-compare-prototype/tree/alpha
+eleventyComputed:
+  eleventyNavigation:
+    key: find-alpha
+    title: "{{ title }}"
 ---
 
-[Alpha version 4](alpha-version-4)
```
fofr, line 18: Why?

package.json
```diff
@@ -24,7 +24,7 @@
     "build:javascripts": "rollup --config etc/rollup.config.js",
     "build": "npm-run-all --parallel build:*",
     "prewatch": "npm run build",
-    "watch:files": "eleventy --serve",
+    "watch:files": "eleventy --serve --quiet",
```
fofr, line 27: 👌

fofr (approved):
❤️ Love this

Tested locally and couldn't find any issues.

If we were to add new products, or nested pages, is there anything special we'd need to do? Is it worth writing some instructions?

Not keen on the renamed files (alpha-version-4 > version-4) because it breaks the old URLs and there aren't any redirects AFAIK.

app/posts/find-teacher-training/2017-12-01-alpha.md
paulrobertlloyd: Repetition in the URL slug was giving my the jitters, but have reversed these changes.

app/_components/breadcrumbs/template.njk
paulrobertlloyd: This filter performs the following transformations:

`path/to/file.html` → `path/to/file`
`path/to/folder/index.html` → `path/to/folder/`

This is less of an issue having moved some (but not all) pages around.

11ty’s preferences is to have all URLs end with a slash – [for good reasons](https://github.com/11ty/eleventy/issues/213#issuecomment-419672043) – although this does require having file saved out as index.html, which makes me cringe a bit.

If we were to ensure all files are output like this, we could remove this filter. This is something I considered for this PR, but stopped myself from expanding it’s scope.

paulrobertlloyd:
> If we were to add new products, or nested pages, is there anything special we'd need to do? Is it worth writing some instructions?

Good idea! There’s one additional change I want to make, which is moving collection indices back into the posts folder. I think makes more sense, and this change makes that possible.

I’ll add an issue to do that, and document adding a new collection (and then back port those changes to the boilerplate).

PR #142, opened 2020-05-18
Move collection indices into respective collection folders

paulrobertlloyd:
Moves index pages for each collection into parent folder, rather than as a separate collection. I _think_ this might make things easier to understand…


fofr (approved):
Much more intuitive :+1:

PR #143, opened 2020-05-18
added new entry on speeding up the references process

EmmaFrith:
There's a lot of content changes in this category so have limited this entry to a few examples


app/posts/apply-for-teacher-training/user-needs/AN0024.md
```diff
@@ -0,0 +1,7 @@
+---
+title: AN024
+userneed: |
+As a candidate
+I need to get 2 references quickly
+So that I don’t miss out on the course that I want
+---
```
fofr:
```suggestion
userneed: |
  As a candidate
  I need to get 2 references quickly
  So that I don’t miss out on the course that I want
---
```

I think

PR #149, opened 2020-05-29
Structured data for pre-degree qualifications

paulrobertlloyd:


benilovj:
Is it worth documenting how we arrived at the underlying data that powers the design? That feels quite important for this feature.

paulrobertlloyd:
@benilovj Added our analysis of the different qualifications and their breakdown, and a link to the spreadsheet we’ve been using. Let me know if you think this needs further detail.

PR #159, opened 2020-06-05
Displaying a safeguarding disclosure

laurahermionetennant92:


app/posts/manage-teacher-training-applications/2020-06-05-view-safeguarding-information.md
```diff
@@ -0,0 +1,42 @@
...
+}) }}
+
+
+2. Users without permission to view sensitive material
+{{ appFigure({
+  image: {
+    file: "for-users-without-permission-to-view-safeguarding-information.png"
+  }
+
+}]
```
fofr:
```suggestion
```
I think this line is breaking the preview app.

app/posts/manage-teacher-training-applications/2020-06-05-view-safeguarding-information.md
```diff
@@ -0,0 +1,42 @@
...
+description: The way we show sensitive material to users differs, depending on whether or not they have permission to view
+date: 2020-06-05
+---
+
+{% from "figure/macro.njk" import appFigure with context %}
+
+
+## Hypotheses
+
+On our provider-facing application form, we created detail components so providers can choose to view our guidance to candidates.
```
adamsilver: The hypothesis structure we're following is as follows:

1. Blurb about the current situation and what's wrong with it
2. If we do X then Y will happen

You could change it like this:

```
Currently, no users can see safeguarding information on an application. But as we're introducing the permission to see safeguarding information we want to let those users know there's information to see.

But we don't want to risk colleagues seeing the information accidentally with an ‘over the shoulder’ attack.

We also want to make sure users without the safeguarding permission know why they can't see it so they can request permission or follow up with the person with safeguarding information as necessary.

If the safeguarding section is disclosed within a collapsible panel
Then users with safeguarding permissions can reveal the information safely

If the safeguarding section explains why it's hidden for users without the permission
Then they'll be able to request permission or follow up with the user who has access
```

adamsilver (changes requested):

PR #166, opened 2020-06-16
Combine product and collection layouts, remove BAT adaptions

paulrobertlloyd:
This PR makes a number of changes which are primarily aimed at making setting up [a new design history site](https://github.com/fofr/govuk-design-history) easier. I’m making the changes on this site first because it’s easier to test how this might work in practice.

Changes include:

* Default `app.url` to `'/'`, to save needing to change that value in `_data/app.json`.
* Use `home` as a more obvious navigation key for the home page.
* Combine the _very_ similar `product` and `collection` layouts into a `collection` layout. If `eleventyNavigation.key == "home"`, use the `app-masthead--large` modifier, increasing the size of the title and description in the masthead.
* The collection layout shows a list of posts as defined in a page’s `pagination` value as it did before. However, if that value _isn’t_ present **and** `eleventyNavigation.key == "home"`, display a list of sections instead. In short, this means that a home page will show all posts on a site by default. However, if pagination is not set, a list of sections – i.e. pages whose parent is set to `home` – is shown instead.
* The labels ‘Posts’ and ‘Sections’ can be customised. For this site, I’ve changed ‘Sections’ to ‘Services’.
* Removed the places where we had divided our 5 services up between Find and Apply. This should reduce the number of places where our code diverges from the new boilerplate/starter site, which we should try to avoid where possible.
* Rather than set `titleHtml` on the homepage, uses a new `widont` filter to add a non-breaking space between the last two words of a string (and applies this to all page titles).


app/_layouts/collection.njk
```diff
@@ -33,33 +34,46 @@
...
-        </section>
-
-        {% if related %}
-        <div class="govuk-grid-column-one-third">
-          {{ appRelated({
-            title: related.title or "Related links",
-            items: related.items
-          }) }}
-        </div>
+        {% if pagination.items %}
```
fofr: I wonder if this forking logic needs an explanatory comment.

eg Show a list of items if using a collection with pagination, otherwise show sections, or something.

fofr: Or use of a temporary variable, `{% if listPosts %}{% elif listSections %}{% endif %}`

app/sitemap.njk
```diff
@@ -3,15 +3,10 @@ layout: page
...
 permalink: "/sitemap.html"
 ---
-{% for item in collections.all | eleventyNavigation("app") %}
-  {% if loop.index == 1 %}
-    <h2 class="govuk-heading-l">Find</h2>
-  {% elseif loop.index == 3 %}
-    <h2 class="govuk-heading-l">Apply</h2>
-  {% endif %}
-  <h3 class="govuk-heading-m govuk-!-margin-bottom-1">
+{% for item in collections.all | eleventyNavigation("home") %}
```
fofr, line 6: What does this bit: `eleventyNavigation("home")` do here?

lib/filters/widont.js
```diff
@@ -0,0 +1,9 @@
+/**
+ * Prevent widows
+ */
+module.exports = string => {
+  const str = String(string);
+  const replaceChar = '\xA0';
```
fofr, line 6:
```suggestion
  const nonBreakingSpaceChar = '\xA0';
```

lib/filters/widont.js
```diff
@@ -0,0 +1,9 @@
...
+  const widontRegex = new RegExp(/\s+([^\s]*)\s*$/)
+  return str.replace(widontRegex, `${replaceChar}$1`)
+}
```
fofr:
```suggestion
module.exports = string => {
  const str = String(string);
  const nonBreakingSpaceChar = '\xA0';
  const widontRegex = new RegExp(/\s+([^\s]*)\s*$/)
  return str.replace(widontRegex, `${nonBreakingSpaceChar}$1`)
}
```

fofr (approved):
Love this. Just a couple of comments to make code easier to understand.

app/sitemap.njk
paulrobertlloyd: It does a couple of things. First, it only shows pages below `home` (it doesn’t include a link to the homepage in the list of pages). It also allows us to use a different presentation for the top level items to any sub pages they may have, which appear nested

Screenshot 2020-06-17 at 10 57 09

Screenshot 2020-06-17 at 11 00 25

.

PR #169, opened 2020-06-18
Refactor user needs

paulrobertlloyd:
Started of by realising we could compute a need’s title from its fileSlug… and then I kinda went to town! 🤪

## Layouts
* Remove separate `user-need` layout and use `post` layout for user needs instead. Related posts now show in the sidebar, like elsewhere on the site.
* Update `user-needs` layout so that in inherits from `page` layout. Remove references to unused macros.

## Content and data
* Compute a user need’s title from its file slug (and prefix with `Need `). This saves us having repeat this information twice (once in the file name, and again in front matter)
* Update all needs to move need out from front matter and into post content. Most needs now don’t require any front matter data.
* Use `eleventyComputed` to generate `related` posts, rather than calculate this in the template.
* Update a post that was missing `tags` that prevented the related need linking back to it.

## Components
* Update `appUserNeed` to be an article with a uniquely named title that is linked. Capitalise status tag text. Refer to `item.data.content` instead of `item.data.userneed`. Adjust styles so that title appears below need and the entire card is selectable (darkening slightly on hover).
* Update `appArticle` component so that it can show a tag in its header. This seems like a thing you might want to do separate from this particular feature (see this [example of a tag shown on an article on the design system website](https://design-system.service.gov.uk/components/accordion/)). Currently have this set up so that the component accepts a `tag` parameter, and in our layouts we give this a post’s `status` value.

### User need component before

User need component before

### User need component after

User need component after

### User need layout before

User need layout before

### User need layout after

User need layout after


adamsilver (approved):
This is great, approved.

Couple things to consider:

1. It's no longer obvious that you can click the the user need card until you hover.

2. Putting the posts that relate to the need on the right hand side could be harder to spot. Also there could be a load of related posts.

fofr:
❤️

adamsilver:
Spotted this – is this a bug:

image

paulrobertlloyd:
@adamsilver:

* Returned underline links on user needs card
* Fixed disappearing text on user need card within a post

and also fixed appearance of parent section in search results (which is broken currently).

paulrobertlloyd:
As to related links, I agree, we might want to visit this presentation later. However, for now, most needs have one or two related posts, so I think we can stick with the existing pattern for now.

PR #171, opened 2020-06-19
Adding a note for the first time

adamsilver:


fofr:
Can we include a bit about why we are also hiding notes from DfE, the provider needs around that, and link to the research? (Or has this been covered elsewhere?)

fofr:
If they don't know how notes work, are they likely to try adding one in the first place?

laurahermionetennant92:
Hi @adamsilver got confused and thought Paul was referring to the prototype not the design history. I do have a tweak to make to that but not this entry.

PR #182, opened 2020-07-14
Trello-powered user needs (proof of concept)

paulrobertlloyd:
Using Trello’s REST API, I wanted to see if was possible to use the user need boards to power the needs we are using on the design history site (thus, making it possible to manage needs in one place, not two).

This PR demonstrates that we can do this, while maintaining all the features we have currently (need embeds, posts related to a specific need, etc):

* [Candidate needs](https://deploy-preview-182--bat-design-history.netlify.app/needs/candidate)
* [Provider needs](https://deploy-preview-182--bat-design-history.netlify.app/needs/provider)
* [Candidate need page](https://deploy-preview-182--bat-design-history.netlify.app/needs/o3kDbr6l)
* [Post with user need embed](https://deploy-preview-182--bat-design-history.netlify.app/apply-for-teacher-training/international-candidates/#user-need)

If we think this is a good idea, next steps would be to:
* Decide if some cards should not be pulled through to the design history site (we can filter cards by tag if necessary)
* Review any needs that are on the design history site but not on the Trello boards, and consolidate them all there
* Review the Trello cards so that they display correctly/better on the design history site (there’s a bit of a mix, especially for provider needs)
* Update existing posts so they point to the Trello based need pages
* Remove the old needs implementation

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -24,6 +24,7 @@ module.exports = function (eleventyConfig) {
   eleventyConfig.addFilter('fixed', require('./lib/filters/fixed'))
   eleventyConfig.addFilter('includes', require('./lib/filters/includes'))
   eleventyConfig.addFilter('markdown', require('./lib/filters/markdown'))
+  eleventyConfig.addFilter('needs', require('./lib/filters/needs'))
   eleventyConfig.addFilter('notifyPlaceholders', require('./lib/filters/notify-placeholders'))
   eleventyConfig.addFilter('pretty', require('./lib/filters/pretty'))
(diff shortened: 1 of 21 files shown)


simonwhatley:
What happens if Trello is ‘down‘ which is somewhat common? Would the design history site cache the information?

paulrobertlloyd:
The site serves static files, so Trello going down wouldn’t effect the live site. This uses 11ty’s [eleventy-cache-assets](https://github.com/11ty/eleventy-cache-assets) plugin, so regenerating the site shouldn’t be a problem either.

PR #189, opened 2020-07-24
Showing international candidate and structured qualifications data

paulrobertlloyd:
Also updated prose styles to:

* Move heading permalink to end of heading. In bulleted lists, the `#` character overlaps the bullet.
* Tweak bullet styles so that they cascade down through to nested lists.


paulrobertlloyd:
There’s still a bit to be added regarding guidance from @laurahermionetennant92; have but a placeholder here in `marked` text.

paulrobertlloyd:
@laurahermionetennant92 will add anything regarding guidance in either an update to this post, or in a separate post.

PR #214, opened 2020-08-19
Move anchors in screenshots and figure to right of heading

paulrobertlloyd:


fofr:
Can we do something to prevent the `#` getting orphaned on a new line?

PR #217, opened 2020-08-19
Prevent heading anchors from being orphaned

paulrobertlloyd:
Narrow viewport:
Screenshot 2020-08-19 at 23 19 39

Slightly less narrow viewport:
Screenshot 2020-08-19 at 23 19 51


fofr (approved):
Nice fix!

PR #231, opened 2020-09-16
Initial DTTP entry added

jesseyuen:
I wanted to capture the screenshots from the DTTP flow we're redesigning in Register for comparison. We can revist this post at a later date if/when we want to add more content.


paulrobertlloyd:
Thanks @jesseyuen, my eyes will never un-see this 😭

jesseyuen:
> Thanks @jesseyuen, my eyes will never un-see this 😭

You can show your thanks with an appproval @paulrobertlloyd!

paulrobertlloyd (approved):

PR #233, opened 2020-09-23
Update images for international and structured qualifications in manage

paulrobertlloyd:
I mean, could write a little post about the changes here (showing labels in qualification cards), but change is minimal and has shipped.


fofr:
General gist is that subheadings were needed, I gather?

paulrobertlloyd:
Added labels within the qualification cards, yeah. Turns out ‘89’ or 7.5 doesn’t mean anything if you don’t say its a grade. Who’d have thunk it?

PR #239, opened 2020-09-29
Asking for degree grades

paulrobertlloyd:


app/posts/apply-for-teacher-training/2020-09-29-degree-grades.md
```diff
@@ -0,0 +1,95 @@
...
+title: Asking for degree grades
+description: Making it easier to enter predicted and international degree grades.
+date: 2020-09-29
+---
+
+Having introduced [structured fields for UK degrees](/apply-for-teacher-training/structured-data-for-degrees/) and allowing candidates to enter [international degrees](/apply-for-teacher-training/international-candidates/#international-qualifications), we reviewed the degree section and found some issues with how we ask for grades.
+
+## Predicted grades
+
+Currently we ask candidates to provide a degree grade. If they haven not completed their degree yet, we ask for a predicted grade:
```
adamsilver:
```suggestion
Currently we ask candidates to provide a degree grade. If they have not completed their degree yet, we ask for a predicted grade:
```

adamsilver:
Did you consider asking a branching question first: `Do you have your degree or are you still studying?`

EmmaFrith:
there appears to be a spacing issue between these 2 lines:

---

Candidates might infer from this that not having a degree yet is considered a lesser option.

We’re effectively asking 2 different questions at the same time:

---

You could also add 'also' into the second line to improve flow

paulrobertlloyd:
@EmmaFrith Yeah, this tripped me up too. It’s because the first line is a caption on image (which looks the same as a paragraph in normal running prose. Annoying, something to fix.

EmmaFrith:
Could we perhaps be a bit more specific on the second hypothesis?

e.g. change this

---
If we ask if you have completed your degree in a separate question
Then we can give guidance at a more relevant time in the user journey

---

to

---

If we ask the 2 different questions on 2 separate pages
Then any guidance relevant to the particular question will be more visible 

--- 

you could even link to the section/screenshot that demonstrates how we implemented this

EmmaFrith:
it may be better to show this screenshot with 'other' selected, so that it's clearer what the surrounding copy means 
https://deploy-preview-239--bat-design-history.netlify.app/images/apply-for-teacher-training/degree-grades/what-grade-is-your-international-degree.png

PR #264, opened 2020-11-06
Making a decision iteration

adamsilver:


paulrobertlloyd:
On the interviews list, is the full address of each location needed given:

* users may already know the addresses of different sites
* it's not important information at this stage
* might be repetitive
* takes up a lot of space

adamsilver:
@paulrobertlloyd that's a different PR but yep, going to see what can be done there. Thanks

PR #273, opened 2020-12-01
Notify placeholder syntax styling for email previews 

paulrobertlloyd:
Allows you to use Notify email syntax for placeholders and conditionals:

`((this is a placeholder))`

`((this is a conditional??thing that only shows if true))`

I experimented with using this on an earlier post, but ended up adding HTML manually. This PR adds a Nunjucks filter to automate this process, and makes this formatting available to the email preview component, `appEmail`. I have also updated previous email templates to use this syntax too.

## Example

The following text:

```
It’s easy to apply again. We’ve saved your last application, so all you
have to do is make any changes and submit.

((has_feedback??Review your feedback and apply again:))

((no_feedback??Apply again:))

((dashboard_url))
```

Appears highlighted as follows:

Screenshot 2020-12-01 at 01 12 22

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -24,6 +24,7 @@ module.exports = function (eleventyConfig) {
   eleventyConfig.addFilter('fixed', require('./lib/filters/fixed'))
   eleventyConfig.addFilter('includes', require('./lib/filters/includes'))
   eleventyConfig.addFilter('markdown', require('./lib/filters/markdown'))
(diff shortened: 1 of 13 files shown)


jesseyuen (approved):
This is great man. Timely too as I need to add an entry on the Publish updated Notifications.

adamsilver:
This is awesome – would it be possible to disable markdown lint inside the email template do you think?

paulrobertlloyd:
@adamsilver I don’t think putting the disable lint rules in the component will work. I’ll open a separate PR to remove/adjust the over zealous rules.

PR #276, opened 2020-12-01
Review Markdown linting

paulrobertlloyd:
* Disable `MD024 - Multiple headings with the same content`
* Move email content into variable blocks, so markdown-lint only needs to run once on a page, not again within `appEmail` component content


adamsilver (approved):
My hero

PR #280, opened 2020-12-08
Blog post for picking publish courses

edwardhorsford:


app/posts/register-trainee-teachers/2020-12-08-picking-publish-courses.md
```diff
@@ -0,0 +1,39 @@
+---
+title: Picking courses from Publish
+description:
+date: 2020-12-08
```
jesseyuen: The date should be closer to when the work was done (or researched) not when the post was added so it fits in chronologically.

jesseyuen (approved):

PR #281, opened 2020-12-08
Added missing email for reasons for rejection

adamsilver:


johndoates (approved):
The email looks good to me.

PR #284, opened 2020-12-09
Improvements to disability, safeguarding and interview sections

adamsilver:


app/posts/manage-teacher-training-applications/2020-12-09-improvements-presenting-disability-safeguarding-interview-information.md
```diff
@@ -0,0 +1,136 @@
...
+- the Details component draws attention and requires a click to view a tiny amount of information
+- a box is too prominent and looks messy inside a table
+
+So we decided to use a combination of grey text and a prefix of ‘Unavailable:’.
+
+## Risks
+
+As we’ve worked on this, we’ve discussed the fact that we’re not storing exactly what the user was asked and what answers they could choose from.
+
+Since the candidate service is still being iterated, this means that we risk the chance of misrepresenting a candidate’s answers over time. The answer will remain the same, but the question it appears to be answering will change.
```
simonwhatley, line 119: There should always be a direct relationship between the question being asked and the answer given and this relationship should be immutable. This avoids the question being restated but the answer remaining the same.

Each application would have a set of questions, guidance and answers. Year on year any of these could change. A minor change (ie the premise remains the same) to the question would not require a new question; a major change will.

simonwhatley (approved):

app/posts/manage-teacher-training-applications/2020-12-09-improvements-presenting-disability-safeguarding-interview-information.md
frankieroberto: I feel like the difficulty / ambiguity here is when exactly does "the premise remain the same"?

For instance, we’re proposing changing "Start date" to "When did you start this job?" and "End date (leave blank if this is your current role)" to "Are you still working in this job?" and "When did you leave this job?". Are these close enough to be the same, or not?

PR #305, opened 2021-01-26
Add post: show reasons why an offer is withdrawn

adamsilver:


frankieroberto (changes requested):
@adamsilver could you add some realistic content to the screenshot, rather than lorem ipsum? Would be helpful for us to understand the kind of reasons why a withdrawal might happen, and expected length of explanation.

Issue #306, opened 2021-01-28
Remove crown, crest and New Transport font


simonwhatley:
The design history isn't on a GOV.UK domain so should not use the crown and crest logos, nor the New Transport font.

paulrobertlloyd:
This has come up a few times. I was wondering if we could use the patterns that the Home Office has created, which are designed for internal tools: https://design.homeoffice.gov.uk. I did briefly look at this, forget what issues prevented me from continuing.

paulrobertlloyd:
(But yes, in the interim remove the crown and font)

fofr:
Let's get it hosted on `.education.gov.uk`

frankieroberto:
We can close this now as it's hosted on a `*.gov.uk` domain... 🙌

paulrobertlloyd:
Sweet URL! (Though now have to update my UX London slides…)

PR #307, opened 2021-01-28
WIP: Replace GOV.UK branding

paulrobertlloyd:
In response #306, thought about using Home Office frontend kit, but it’s not very easy to use, unfortunately (makes lots of assumptions). But looking at it, can probably change much of the design by overriding some GOV.UK Frontend defaults.

I dunno, could just remove the crown and font, but this might work too. Thoughts?

### DIFF
--- a/app/_components/_all.scss
+++ b/app/_components/_all.scss
@@ -6,6 +6,7 @@
 @import "email/email";
 @import "embed/embed";
 @import "figure/figure";
+@import "footer/footer";
 @import "gallery/gallery";
 @import "header/header";
 @import "masthead/masthead";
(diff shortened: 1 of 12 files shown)


ghost:
Great stuff, when this WiP is finished we're going to have steal this...

paulrobertlloyd:
@Gkionis No guarantees this will ship, just a proof of concept. Why is this change important to you out of interest?

ghost:
We could use an internal branding with the DfE 'unicorn and lion' insignia, senior stakeholders don't like the Gov.uk branding. And I don't believe the words I am typing but I agree with them on this.

PR #318, opened 2021-02-17
Make heading anchors only visible on hover

paulrobertlloyd:
* Add initial letters to each item in glossary to make it easier to browse
* Only show header anchors upon hovering over a heading

### DIFF
--- a/app/_components/prose/_prose.scss
+++ b/app/_components/prose/_prose.scss
@@ -178,15 +178,22 @@
   }
 
   .header-anchor {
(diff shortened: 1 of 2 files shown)


frankieroberto (approved):
Like it. 👍 

I never knew this page existed though, as it's only linked to from the footer (and no-one ever looks there... 😂 ). I might open a PR with an idea for linking to it from the homepage.

edwardhorsford (approved):

PR #319, opened 2021-02-18
Link to glossary from homepage

frankieroberto:
The glossary feels a bit hidden in the footer, but is super useful.

I've kept the other links in the footer, as they all link to external websites.

Link and text is hardcoded, but if there’s a way to do this neatly within Eleventy I’m all ears.

### DIFF
--- a/app/_layouts/base.njk
+++ b/app/_layouts/base.njk
@@ -52,9 +52,6 @@
         }, {
           text: "Wiki",
           href: "https://dfedigital.atlassian.net/wiki/spaces/BaT/overview"
-        }, {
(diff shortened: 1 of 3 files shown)


paulrobertlloyd:
LOVE. IT!

(I’ve been thinking most – if not all – of the links in the footer can be got rid of, as non relevant or used anymore)

paulrobertlloyd (approved):

frankieroberto:
@paulrobertlloyd heh, I did wonder whether the "Wiki" (which I can't access) was still being used at all.

PR #321, opened 2021-02-19
Bump node to version 15

frankieroberto:
Only version 15+ of Node.js is supported on M1 (ARM) Macs, and I have a new MacBook so... 😄


frankieroberto:
The Netlify `node_modules` cache needs to be purged, I think, to make all the checks pass.

PR #322, opened 2021-02-20
Add mission patch archive

paulrobertlloyd:
Adds a page to the design history (linked to from the home page) displaying [all the mission patches we have designed for different services, milestones and projects](https://deploy-preview-322--bat-design-history.netlify.app/mission-patches).

There are still a few gaps that need filling in before we can publish this page:

* [x] @jesseyuen - can you add details behind the choice of an Octopus, and a bit of background behind the mission patch for the private beta?
* [x] @jesseyuen - can you add the mission patch you created for the first recruitment cycle (and the first appearance of the space canary)
* [x] @jesseyuen - can you add a bit of details about why we use a fox for our allocations projects?
* [x] @paulrobertlloyd – I’ll find out why we used a lobster for our data discovery
* [x] @fofr - Can you provide some background on why Find uses a canary, and possibly dig out the first mission patch you created for this service?


fofr:
Canary story:

Miners used to carry canaries down into the coal mines as an early warning system, if the canaries died then the carbon monoxide levels in the mines were too high, and the miners could get out before they also died.

When we launched Publish we weren’t sure how much our recent integration with DfE Sign-in would work, and our first few providers were our canaries, to test they could access and navigate the service. Our first versions failed and we had many dead canaries until we got the process right, then we rolled it out to everyone.

fofr (approved):

PR #354, opened 2021-04-21
Add visa sponsorship reason for rejection

simonwhatley:


app/posts/manage-teacher-training-applications/2021-04-19-reason-for-rejection-cannot-sponsor-visa.md
```diff
@@ -0,0 +1,70 @@
...
+
+15% of international candidates’ right to work or study in the UK is unknown (for example, marked as ‘not yet, or not sure’).
+
+For these candidates, one of the reasons their application can be rejected is the provider cannot sponsor their visa.
+
+## What we did and why
+
+To provide consistent and useful feedback to candidates, we added the question “Was it because you cannot sponsor the candidate’s visa?” to the reasons for rejection.
+
+The question is only shown when the candidate is not a British or Irish citizen, they are ‘international’.
```
frankieroberto: @simonwhatley  Would it make sense to only show this question if the candidate is not a British or Irish citizen AND they answered "No or not yet" to "Do you already have the right to work or study in the UK?" – as it doesn’t really make sense as a rejection reason if they already have the right to work or study here (eg an EU citizen with settled status).

simonwhatley: Yep, that makes sense. We're going to consider reducing the range of candidates for which we show the option after we've learnt more about the question's use.

The right to work or study question is interesting in so much as their right may be reported incorrectly or may expire prior to the start of the course, therefore requiring a new visa.

One of the reasons we're leaving the question open to a broader range of candidates is so that the provider can also learn of its existence.

PR #367, opened 2021-05-10
Rollover 2020

jesseyuen:
Rollover 2020 post added


paulrobertlloyd:
Could we show the emails inline, using the `appEmail` helper?

paulrobertlloyd:
It might annoy me if I wanted to find out this information later, and couldn’t because emails were deleted from Notify.

edwardhorsford (approved):

PR #403, opened 2021-07-27
Simpler syntax for figures and screenshots (proof of concept)

paulrobertlloyd:
We should try to make it easier to:

* Add single images to posts
* Create posts containing a selection of screenshots

This PR is a proof of concept of how we can achieve this.

## Simpler figures

Currently, to add a `figure` to a post, you have to use the following Nunjucks component:

```njk
{% from "figure/macro.njk" import appFigure with context %}
{{ appFigure({
  image: {
    file: "adding-users-01.png",
    alt: "Provider’s user list page"
  },
  caption: "Figure 1: Provider’s user list page"
}) }}
```

However, using the [`markdown-it-image-figures`](https://github.com/Antonio-Laguna/markdown-it-image-figures) plugin, we can leverage the [little used Markdown syntax](https://www.markdownguide.org/basic-syntax/#images-1) that allows you to add a `title` attribute to an image, but instead use that for the `figcaption`:

```markdown
![Provider’s user list page](adding-users-01.png "Figure 1: Provider’s user list page")
```

A few notes:
* We need to move the CSS for figures into `.app-prose` to ensure all figures get the same styles
* To save having to write the full file path, updated `addPassthroughCopy` to copy images into the same output directory as the post
* If we use this for all figures on the site, we can remove the `appFigure` (and `appImage`) components.

## Simpler screenshots

Currently, to add a set of screenshots to a post, you have to use the following Nunjucks component:

```njk
{% from "screenshots/macro.njk" import appScreenshots with context %}
{{ appScreenshots({
  items: [{
    text: "Applications",
    img: { src: "01-applications.png" }
  }, {
    text: "Unsubmitted application",
    img: { src: "02-unsubmitted-application.png" }
  }]
}) }}
```

We can simplify this so that we render screenshots at the bottom of the page if a post has a `screenshots` in its frontmatter:

```yaml
screenshots:
  items:
    - title: Applications
      src: 01-applications.png
    - title: Unsubmitted application
      src: 02-unsubmitted-application.png
```

You can also provide multiple sets of screenshots:

```yaml
screenshots:
- title: First set of screenshots
  items:
    - title: Applications
      src: 01-applications.png
    - title: Unsubmitted application
      src: 02-unsubmitted-application.png
- title: Second set of screenshots
  items:
    - title: Applications
      src: 01-applications.png
    - title: Unsubmitted application
      src: 02-unsubmitted-application.png
```

You can also use a simplified list of screenshot titles (this works if image filenames use slugified titles):

```yaml
[...]

PR #406, opened 2021-07-29
Use relative image paths

paulrobertlloyd:
In preparation for simplifying figures and screenshots (see #403), update where we serve images from:

* When generating the site, copy files from the `images` folder, and save alongside post HTML. This means that, in most cases, we can continue to only call an image’s filename, but without needing to do any workarounds in the different components to get the right path
* Remove `appImage` component.
  * When used in Markdown files, use Markdown image syntax instead
  * When used in `appFigure` component, use HTML instead
* Remove `path` param on `appFigure` component, provide full path if necessary in `file` param (the `appFigure` component will be removed in a later PR)
* Update places where we reference images being in the `images` folder to point to new image locations

### DIFF
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -35,10 +35,8 @@ module.exports = function (eleventyConfig) {
 
   // Passthrough
   eleventyConfig.addPassthroughCopy('./app/documents')
-  eleventyConfig.addPassthroughCopy('./app/images')
-  eleventyConfig.addPassthroughCopy({
-    'node_modules/govuk-frontend/govuk/assets': 'assets'
-  })
+  eleventyConfig.addPassthroughCopy({'./app/images': '.'})
+  eleventyConfig.addPassthroughCopy({'node_modules/govuk-frontend/govuk/assets': 'assets'})
 
   // Enable data deep merge
   eleventyConfig.setDataDeepMerge(true)
--- a/.eleventy.js
+++ b/.eleventy.js
@@ -35,8 +35,8 @@ module.exports = function (eleventyConfig) {
 
   // Passthrough
   eleventyConfig.addPassthroughCopy('./app/documents')
-  eleventyConfig.addPassthroughCopy({'./app/images': '.'})
-  eleventyConfig.addPassthroughCopy({'node_modules/govuk-frontend/govuk/assets': 'assets'})
+  eleventyConfig.addPassthroughCopy({ './app/images': '.' })
+  eleventyConfig.addPassthroughCopy({ 'node_modules/govuk-frontend/govuk/assets': 'assets' })
 
   // Enable data deep merge
   eleventyConfig.setDataDeepMerge(true)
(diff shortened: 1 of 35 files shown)


frankieroberto:
Could we avoid the copying images step altogether, I wonder, and reference them directly from the markdown in their actual location? Perhaps even do away with the separate images folder and have an `images` subfolder of `posts/apply-for-teacher-training/`?

The copying step has thrown me a few times – and referencing them in their original location would allow GitHub markdown previews to work?  Would also avoid having to match the post slug with the images subfolder name (minus the date), which isn't intuitive.

simonwhatley:
My preference is to have the images and markdown in the same folder for each post. 

So instead of having content in two places (`images` and `posts`):

```
/images/service-name/post-name/
  - image-01.png
  - image-02.png
  - image-03.png
/posts/service-name/
  - 2020-07-30-post-name.md
```

We would have content in one place in the `posts` folder:

```
/posts/service-name/post-name/
  - image-01.png
  - image-02.png
  - image-03.png
  - index.md
```

The latter simplifies the structure for other types of media such as documents and videos.

frankieroberto:
> My preference is to have the images and markdown in the same folder for each post.

I like this idea too – saves having to match the folder names (which can be a pain, if you rename one).

paulrobertlloyd:
Only downside with using folders is that posts are no longer ordered by date. But benefits are that only need to set the publish date in one place, and maybe alphabetically ordered folders are easier to use when looking for a previously published post. So, if we’re all agreed, let’s move to having posts and images together in the same folder. I’ll update this PR.

edwardhorsford:
I'd be really keen on everything about a post in one - and using dated folders.

I've seen several cases where the date in the filename doesn't match the date in the post - would be good to clean this up.

frankieroberto:
@edwardhorsford:

> I've seen several cases where the date in the filename doesn't match the date in the post - would be good to clean this up.

Isn’t this an argument against having dates in the folders, as it means the date is in two places? Or are you suggesting _only_ having the date in the folder?

Personally, I’m never sure what to put the date as, as it can take a few days to write and publish a post (is it meant to be publish date, or just whenever you started writing it?)

edwardhorsford:
> @edwardhorsford:
> 
> > I've seen several cases where the date in the filename doesn't match the date in the post - would be good to clean this up.
> 
> Isn’t this an argument against having dates in the folders, as it means the date is in two places? Or are you suggesting _only_ having the date in the folder?
> 
> Personally, I’m never sure what to put the date as, as it can take a few days to write and publish a post (is it meant to be publish date, or just whenever you started writing it?)

I'm suggesting there should be only one date, wherever it is.

Personally I'd be pro publish date, but so far in register they've been backdated to when we were doing the work.

paulrobertlloyd:
I originally tried dated folders, but was having difficulty getting 11ty to play ball. And right now, the navigation plugin is making things even more difficult. Fast falling out of love with 11ty, if I’m honest, and certainly the navigation plugin. So, so fiddly. 😡

paulrobertlloyd:
While I agree with moving images to live in the same folders as posts, doing so is causing issues with the navigation plugin. I’m going to merge this in, create PRs for the other simplification changes, and then come back to folder reorganisation after all that’s done. (Eleventy is making me pull my hair out…)

PR #412, opened 2021-08-03
Use Markdown for figures and remove `appFigure` component

paulrobertlloyd:
* Remove `appImage` component (thought I’d done this in #406, but evidently not)
* Add `markdown-it-image-figures` plugin
* Apply `.app-figure` styling to all `figure` elements within `.app-prose`
* Replace `appFigure` component in `appScreenshots` component with plain HTML
* Replace `appFigure` component in `appGallery` component with plain HTML
* Replace all occurrences of `appFigure` component in posts with simpler Markdown image syntax
* Remove `appFigure` component
* Update image filenames for a number of Register posts so that they use kebab case

## How to include images in posts

With this change, we can now use the straightforward Markdown image syntax. So instead of this: 

```njk
{% from "figure/macro.njk" import appFigure with context %}
{{ appFigure({
  image: {
    file: "adding-users-01.png",
    alt: "Provider’s user list page"
  },
  caption: "Figure 1: Provider’s user list page"
}) }}
```

we can now write this:

```markdown
![Provider’s user list page](adding-users-01.png "Figure 1: Provider’s user list page")
```

This change means 1 line can now be written instead of 5 or 7. This might go someway to explain why this PR removes ~1700 lines of code 🤯

## A note about alternative text

A side-effect of doing this is that Markdown Lint will now give a warning about any image without alternative text.

**We have 164 images on the site that don’t have alt text!**

It’d be nice if we could at some point revisit these images to make sure they are properly described.


paulrobertlloyd:
Hmm, I guess to have checks pass, I’ll have to add an exception to Markdown Lint not to warn on images without alt text…

paulrobertlloyd:
So, guess who added alt text to all 164 images… all of which were for the _Manage_ service, I should add 😅

PR #439, opened 2021-08-19
Auto linter

frankieroberto:
We have a "linter" task set up which checks for some minor code style things in the javascript and markdown files. Both of those tools have an "auto-fix" mode, which attempts to automatically fix as many things as possible (some require human intervention).

This PR lets you easily run them both in that mode by running:

```bash
npm run lint:fix
```

In addition, for any Pull Requests opened on GitHub, GitHub will run the linters in that mode, and then will push any fixes. It will still alert you if there are any remaining issues which couldn’t be fixed automatically.

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -13,5 +13,14 @@ jobs:
     - uses: actions/checkout@v2
     - name: Install packages
       run: npm install
-    - name: Run linters
+    - name: Run linters with autofix
+      run: npm run lint:fix
+    - name: Push any fixes
+      run: |
+        git config --global user.email "lint-bot@example.com"
+        git config --global user.name "lint-bot"
+        git add --all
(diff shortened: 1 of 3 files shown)


paulrobertlloyd:
:heart: this! Not sure we’ve got linting on the main GOV.UK Design History, repo, otherwise I’d suggest adding this to that project too. Maybe we can add to the Apply and Find prototypes?

frankieroberto:
> Not sure we’ve got linting on the main GOV.UK Design History, repo, otherwise I’d suggest adding this to that project too. 

Don't think I even know about that project. Where is it?

> Maybe we can add to the Apply and Find prototypes?
👍 

The one downside is that if your git habit is to continually rebase your local branch and force push, you might get conflicts. (Personally I consider force-pushing bad practice, but everyone has a different view on this).

PR #443, opened 2021-08-20
New subject flow on Find

frankieroberto:
➡️ [Preview](https://deploy-preview-443--bat-design-history.netlify.app/find-teacher-training/new-subject-flow/)


app/posts/find-teacher-training/2021-08-19-new-subject-flow.md
```diff
@@ -0,0 +1,83 @@
...
+title: New subject flow
+description: Why we changed how users select course subjects
+date: 2021-08-19
+related:
+  items:
+    - text: Special educational needs (SEN) filter
+      href: /find-teacher-training/sen-filter/
+    - text: SEND survey and updates
+      href: /find-teacher-training/send-survey/
+
```
paulrobertlloyd:
```suggestion
```

app/posts/find-teacher-training/2021-08-19-new-subject-flow.md
```diff
@@ -0,0 +1,83 @@
...
+If a user answers yes, they are shown a list of the specialist subjects. A checkbox at the bottom allows them to also see primary courses without a specialist subject.
+
+![Screenshot of the new page asking users what specialist subjects they are interested in](primary-specialist.png)
+
+## Secondary courses
+
+If someone wants to teach the secondary age group, we show them a list of all the subjects in alphabetical order. Whilst this is a long list, we found in usability tests that users were able to find the subjects they were interested in quickly.
+
+![](secondary.png)
```
paulrobertlloyd: Should we add alt text for this image?

paulrobertlloyd (approved):

EmmaFrith:
"Instead, the simplest initial thing to do is to remove SEND as a checkbox from both the subject pages altogether, and keep it as a separate filter on the results page."

What's the behaviour when it's on the results page instead? Why are users more likely to understand it there?

app/posts/find-teacher-training/2021-08-19-new-subject-flow.md
frankieroberto: Yep, opps - this was to re-test the auto-linter.

frankieroberto:
> "Instead, the simplest initial thing to do is to remove SEND as a checkbox from both the subject pages altogether, and keep it as a separate filter on the results page."
> 
> What's the behaviour when it's on the results page instead? Why are users more likely to understand it there?

The behaviour is the same as it on the live site at the moment - ie checking the box filters the results to only courses with a SEN specialism.

Potentially users are more likely to understand it there as it is more clearly separated from the subject question. They also would see it after already seeing some results (which may include some SEND-specialist courses), which might help. But perhaps it could still be improved. More research needed! 😄

PR #484, opened 2021-11-01
Make website indexable in search engines like Google

adamsilver:
It's difficult to find posts in the design history using the autocomplete search alone.

Enabling the site to be spidered will let users find things via Google, for example.

It'll also let us add a Google search form to the site if we choose to do that later.


frankieroberto (approved):

simonwhatley (changes requested):
Don't delete robots.txt, just replace the content with:

```
User-agent: *
Allow: /
```

We can use the robots file later to point to a sitemap.xml.

simonwhatley (approved):

PR #485, opened 2021-11-01
Add sitemap.xml

simonwhatley:
This is useful for search engine indexing.

### DIFF
--- a/app/sitemap.xml.njk
+++ b/app/sitemap.xml.njk
@@ -1,0 +1,17 @@
+---
+permalink: sitemap.xml
+eleventyExcludeFromCollections: true
+---
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+  {%- for page in collections.all %}
+  {%- if page.data.layout %}
+  <url>
+    <loc>{{ site.url }}{{ page.url }}</loc>
+    <lastMod>{{ page.date | rssDate }}</lastMod>
+    <changefreq>{{ page.sitemap.changefreq | default('monthly') }}</changefreq>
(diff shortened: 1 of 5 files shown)


app/sitemap.xml.njk
```diff
@@ -0,0 +1,17 @@
...
+permalink: sitemap.xml
+eleventyExcludeFromCollections: true
+---
+<?xml version="1.0" encoding="UTF-8"?>
+<urlset xmlns="http://www.sitemaps.org/schemas/sitemap/0.9">
+  {%- for page in collections.all %}
+  {%- if page.data.layout %}
+  <url>
+    <loc>{{ site.url }}{{ page.url }}</loc>
+    <lastMod>{{ page.date | rssDate }}</lastMod>
```
frankieroberto: Looks like this should work, and avoids the need for a plugin:

```suggestion
    <lastMod>{{ page.date.toISOString() }}</lastMod>
```

package.json
```diff
@@ -40,6 +40,7 @@
   "dependencies": {
     "@11ty/eleventy": "^0.12.1",
     "@11ty/eleventy-navigation": "^0.3.2",
+    "@11ty/eleventy-plugin-rss": "^1.1.2",
```
frankieroberto: Might not be needed, if this is just for string formatting? See above.

frankieroberto (approved):
Made a suggestion that avoids the need for the plugin.

Looks like this is currently relying on our manually-set post dates, which might not always be accurate (if we publish or update after setting them)? Perhaps we could switch to using the "Last modified" date from https://www.11ty.dev/docs/dates/ instead somehow?

package.json
simonwhatley: I was going to add some Atom/RSS stuff at a later date, but happy to remove it.

frankieroberto: Remove until we add it?

PR #487, opened 2021-11-04
Remove "– GOV.UK " from page title

frankieroberto:
Should probably remove this, as the site isn't part of GOV.UK?

### DIFF
--- a/app/_layouts/base.njk
+++ b/app/_layouts/base.njk
@@ -14,7 +14,7 @@
 {% endblock %}
 
 {% block pageTitle %}
-  {% if title %}{{ title }} - {% endif %}{{ app.serviceName or app.productName }} - GOV.UK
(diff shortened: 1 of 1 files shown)


app/_layouts/base.njk
```diff
@@ -14,7 +14,7 @@
 {% endblock %}
 
 {% block pageTitle %}
-  {% if title %}{{ title }} - {% endif %}{{ app.serviceName or app.productName }} - GOV.UK
+  {% if title %}{{ title }} - {% endif %}{{ app.serviceName or app.productName }}
```
simonwhatley: Instead of `GOV.UK` should we put `Department for Education`?
```suggestion
  {% if title %}{{ title }} - {% endif %}{{ app.serviceName or app.productName }} - Department for Education
```

frankieroberto: Good idea.

PR #509, opened 2021-12-27
Use govuk-eleventy-plugin

paulrobertlloyd:


simonwhatley:
Whilst abstracting the code is generally good, I think we need to consider whether we should merge this change carefully. It makes it harder for us to optimise the BAT design history. 

A current pain point is the slow building of a page. Another is how we manage images. We can't address these without contributing to the plugin and all the process change that entails.

cc @adamsilver @frankieroberto @edwardhorsford

paulrobertlloyd:
> Whilst abstracting the code is generally good, I think we need to consider whether we should merge this change carefully. It makes it harder for us to optimise the BAT design history.

Up to the team if you want to merge this, of course. Will just point out that most (if not all) functionality provided by the plugin can be overridden. Indeed, the layouts provided by the plugin are superseded in the Design History project (and therefore in  this PR) to support showing screenshots and related items on posts and pages.

> A current pain point is the slow building of a page. Another is how we manage images. We can't address these without contributing to the plugin and all the process change that entails.

This is the configuration the plugin in provides: https://github.com/paulrobertlloyd/govuk-eleventy/blob/main/package/.eleventy.js

Perhaps including Nunjucks in the plugin is overkill. Downside however is that setting up a new project then becomes more cumbersome. I *think* the way 11ty is set up, you could extend/overwrite Nunjucks within this projects own 11ty config. At the moment the plugin allows you add directories for Nunjucks to find additional views and templates. Not sure how you would add other filters… that should be easier than it perhaps is.

Thanks for the feedback!

PR #512, opened 2022-01-07
Add some terms to the glossary

frankieroberto:
Adds some terms to the glossary which are used on the Find and Apply services, and includes some guidance on how they should be used:

* Ancient languages
* DBS Check
* English as a foreign language
* Fitness to teach check
* Modern languages
* Postgraduate diploma in education (PGDE)
* Special educational needs and disability (SEND or SEN)
* UK ENIC

### DIFF
--- a/app/glossary.md
+++ b/app/glossary.md
@@ -24,6 +24,9 @@ Accreditation
 Allocation
 : Permission to run limited courses.
 
+Ancient languages
+: Used to describe teaching the study of languages no longer in widespread spoken usage, such as Latin or Ancient Greek. Prefer using the specific languages studied where possible.
+
 ## B
 
 Becoming a teacher (BAT)
@@ -48,6 +51,9 @@ Cycle
 Database of Qualified Teachers (DQT)
 : A database maintained by the Teacher Regulation Agency. Teachers can sign in to view their teacher record, get copies of QTS and other teaching related qualifications and update their personal and employment details.
 
+DBS Check
+: A service from the Disclosure and Barring Service which checks for any potential criminal record. There are two levels, a ‘basic’ check and an ‘enhanced’ check. Applying for teacher training requires the enhanced check. Generally, the acronym does not need to be spelled out, but the check should be described, for example as a criminal record check.
+
 DTTP
 : [Database of trainee teachers and providers](https://www.gov.uk/guidance/database-of-trainee-teachers-and-providers-dttp).
 
@@ -59,6 +65,9 @@ Early career payments
 English Baccalaureate (EBacc)
 : A combination of subjects (the ‘core’ subjects, academically) that pupils can study at GCSE level. The EBacc is not a qualification in itself. The government assesses school performance based on how well pupils perform in the EBacc subjects. It’s also used as something of a marketing device - pupils are encouraged to choose these subjects.
 
+English as a foreign language
+: Used to describe the use of English by people whose first language or languages are not English. The teaching of this is a specialism sometimes described as ‘teaching English as a foreign language’ (TEFL). The use of English as a foreign language can be assessed through formal assessments such the ‘International English Language Testing System’ (IELTS) and ‘Test of English as a Foreign Language’ (TOEFL).
+
 Early years initial teacher training (EYITT)
 : The equivalent to ITT but specifically for teaching 0 to 5 year olds.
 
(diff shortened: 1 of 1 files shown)


app/glossary.md
```diff
@@ -48,6 +51,9 @@ Cycle
 Database of Qualified Teachers (DQT)
 : A database maintained by the Teacher Regulation Agency. Teachers can sign in to view their teacher record, get copies of QTS and other teaching related qualifications and update their personal and employment details.
 
+DBS Check
+: A service from the Disclosure and Barring Service which checks for any potential criminal record. There are two levels, a ‘basic’ check and an ‘enhanced’ check. Applying for teacher training requires the enhanced check. Generally, the acronym does not need to be spelled out, but the check should be described, for example as a criminal record check.
```
EmmaFrith: I think there's actually 3 levels! https://www.gov.uk/government/organisations/disclosure-and-barring-service/about#dbs-checks

app/glossary.md
```diff
@@ -101,6 +113,9 @@ Lead school
 Marketing name
 : The commonly used name of an educational organisation. This may not be the legal name.
 
+Modern languages
+: This term is sometimes used to differentiate the teaching of languages in everyday use from ‘ancient languages’ such as Latin and Ancient Greek which are no longer widely spoken. It also typically does not include teaching English. Avoid using in non-internal communication.
```
EmmaFrith: don't think we can advise avoiding using 'modern languages' as it's all over course name. we currently also use it on Find search

app/glossary.md
```diff
@@ -122,6 +137,9 @@ Organisation
 Postgraduate certificate in education (PGCE)
 : This is a teaching qualification that - when it’s paired with QTS (qualified teacher status) - allows you to teach in state schools in England, Scotland, Wales and Northern Ireland. It’s also recognised internationally (which is the major advantage of having it, over just QTS). PGCE without QTS qualifies you to teach further education, but not primary or secondary education.
 
+Postgraduate diploma in education (PGDE)
+: Very similar to the PGCE. Generally includes more credits towards a Master’s degree, but is essentially the same.
```
EmmaFrith:
```suggestion
: Very similar to the PGCE. Generally includes more credits towards a Master’s degree.
```

app/glossary.md
```diff
@@ -180,8 +201,11 @@ Teaching School Alliance
 UCAS
 : The Universities and Colleges Admissions Service whose main role is to operate the application process for British universities.
 
+UK ENIC
+: A service which provides impartial judgement on the international qualifications, and how they compare to UK qualifications. Previously known as UK NARIC prior to the UK leaving the European Union. ENIC stands for European Network of Information Centres, but the term does not need to be spelled out. See [UK ENIC](https://www.enic.org.uk) website.
```
EmmaFrith:
```suggestion
: A service which provides impartial judgement on international qualifications, and how they compare to UK qualifications. Previously known as UK NARIC prior to the UK leaving the European Union. ENIC stands for European Network of Information Centres, but the term does not need to be spelled out. See [UK ENIC](https://www.enic.org.uk) website.
```

app/glossary.md
```diff
@@ -180,8 +201,11 @@ Teaching School Alliance
...
+
 UK NARIC
-: All EU and EEA states (as well as some other associated countries) have a designated National Academic Recognition Information Centre that is responsible for recognising international qualifications and skills. The [National Recognition Information Centre for the United Kingdom](https://www.naric.org.uk/) is the UK’s agency. Providers can ask international applicants to provide a [statement of comparability](https://www.naric.org.uk/Qualifications/SOC/Default.aspx) to ensure their qualifications align with UK requirements.
+: Previous name for UK ENIC, whilst the UK was part of the EU.
```
EmmaFrith, line 208: does this line need to be deleted?

app/glossary.md
frankieroberto: Or even 4?!

app/glossary.md
frankieroberto: It should be getting dropped from the course search soon... 😄. Longer term I’d like to review how it's used in course titles too, as it can mean a few different things...

In the meantime maybe we should downgrade the guidance to something like "use the specific languages where possible" or something?

app/glossary.md
frankieroberto: I figured it might might worth keeping it around for a bit longer, as we might still use it in a few places (in case anyone was familiar with the old name)?

app/glossary.md
frankieroberto: Have updated.

app/glossary.md
frankieroberto: I've updated to "Prefer listing the specific languages where possible".

app/glossary.md
EmmaFrith: ah I see, yea fine!

EmmaFrith (approved):

PR #527, opened 2022-01-25
Update glossary

tvararu:
Add CPD and TVS. CPD definition courtesy of @benilovj.

### DIFF
--- a/app/glossary.md
+++ b/app/glossary.md
@@ -45,6 +45,10 @@ These are sums of money available to trainees on (fee-based) courses in certain 
 
 ## C
 
+### Continuous professional development (CPD)
+
+A generic term referring to training of people already qualified and in employment. Avoid using the acronym, or spell it out on first use.
+
 ### Course
 
 A training programme providers create for candidates to apply to.
@@ -246,6 +250,10 @@ Used to predict and manage the ITT market.
 ### Teaching School Alliance
 
 Outstanding schools that work with others to provide high quality training to new and experienced staff. Cannot be lead schools or providers.
+
+### Teaching Vacancies (TVS)
+
+A service for [advertising, finding and applying for teaching jobs](https://teaching-vacancies.service.gov.uk/). Capitalised. Do not use the acronym publicly.
 
 ## U
 


app/glossary.md
```diff
@@ -45,6 +45,10 @@ These are sums of money available to trainees on (fee-based) courses in certain
 
 ## C
 
+### Continuous Professional Development (CPD)
+
+Area around offering training for teachers.
```
frankieroberto: I don't think this term is teacher-specific? And should it be de-capitalised? 🤔 

```suggestion
### Continuous professional development (CPD)

A generic term referring to training of people already qualified and in employment. Avoid using the acronym, or spell it out on first use.
```

app/glossary.md
```diff
@@ -247,6 +251,10 @@ Used to predict and manage the ITT market.
 
 Outstanding schools that work with others to provide high quality training to new and experienced staff. Cannot be lead schools or providers.
 
+### Teaching Vacancies (TVS)
+
+Service to [find and apply for teaching jobs](https://teaching-vacancies.service.gov.uk/).
```
frankieroberto:
```suggestion
A service for [advertising, finding and applying for teaching jobs](https://teaching-vacancies.service.gov.uk/). Capitalised. Do not use the acronym publicly.
```

frankieroberto (changes requested):
Thanks for adding these! We should also add [National professional qualifications (NPQs)](https://www.gov.uk/government/publications/national-professional-qualifications-npqs-reforms/national-professional-qualifications-npqs-reforms) but that could be a separate PR.

The linter doesn't seem to run on forks, but hopefully that'll be fixed in #529.

app/glossary.md
edwardhorsford: Should `vacancies` be capitalised?

frankieroberto: Possibly not, but it is on [the service](https://teaching-vacancies.service.gov.uk/).

tvararu:
Cheers @edwardhorsford and @frankieroberto for the feedback! I didn't have notifications turned on which is why it took me days to commit the suggestions :sweat_smile:

frankieroberto (approved):

PR #564, opened 2022-02-24
Add service map

frankieroberto:
Adds the current service map as a reference, so that you can see how BAT services relate to other teacher services. Also lets people download the PDF.

Probably this could ultimately 'live' somewhere else, as it's broader than just BAT, but in the meantime at least this would let people access it.

I've added an HTML version of the diagram underneath it, for accessibility and so that we can link to the actual services.

➡️ [Preview](https://deploy-preview-564--bat-design-history.netlify.app/service-map/)


app/service-map.md
```diff
@@ -0,0 +1,206 @@
+---
+layout: page
+title: Service map
```
ralph-hawkins: I'd argue this wasn't a 'service map' as most people would expect. Maybe call it the 'Teacher services diagram'?

app/service-map.md
```diff
@@ -0,0 +1,206 @@
...
+description: How the ‘becoming a teacher’ services relate to other services in the teaching journey
+tags: reference
+eleventyComputed:
+  eleventyNavigation:
+    key: "{{ title }}"
+    excerpt: "{{ description }}"
+    parent: home
+---
+
+The 'becoming a teacher' services are part of a broader range of digital services which encompass the whole teaching journey, from initial interest through to retirement.
```
ralph-hawkins, line 13:
```suggestion
The ‘becoming a teacher’ services are part of a broader range of digital services which encompass the whole teaching journey, from initial interest through to retirement.
```

ralph-hawkins (changes requested):
I find the HTML version pretty hard to read — could you use a table or CSS grid to add horizontal alignment?

app/service-map.md
frankieroberto: Map and diagram feel fairly synonymous to me? (eg "tube map" / "mind map"). It's probably technically a diagram, but I think most people call it a map? 

How about "Teacher services map"?

frankieroberto:
> I find the HTML version pretty hard to read — could you use a table or CSS grid to add horizontal alignment?

I started with a table, but then felt that it'd probably need another column of headers to the left, which was getting a bit tight. Also worried that it was abusing tables for layout a bit...

I'll have a look at CSS grids though.

app/service-map.md
ralph-hawkins: Yeah, that works

frankieroberto:
> > I find the HTML version pretty hard to read — could you use a table or CSS grid to add horizontal alignment?
> 
> I started with a table, but then felt that it'd probably need another column of headers to the left, which was getting a bit tight. Also worried that it was abusing tables for layout a bit...
> 
> I'll have a look at CSS grids though.

@ralph-hawkins I've done the simplest thing I could think of and have added bullets to the lists, which hopefully separates them from the lead-in lines a bit more:

Screenshot 2022-02-28 at 16 29 24

ralph-hawkins (approved):

PR #589, opened 2022-04-04
New post: Emailing candidates about upcoming interviews when their course is changed

johndoates:


adamsilver:
I don't think we need to say "which the user belongs to" because it could be the accredited body user who makes the change to another training provider.

PR #623, opened 2022-06-15
Add post for revisiting how we identify current and past trainees

edwardhorsford:


edwardhorsford:
[Design history entry for cohort / helping users distinguish past and current trainees](https://trello.com/c/HQWkHQhy/4190-design-history-entry-for-cohort-helping-users-distinguish-past-and-current-trainees)

PR #721, opened 2022-10-18
Update post: Improving the content on the search results page and course page

simonwhatley:
Removed:

> We changed ‘Degree grade accepted’ to ‘Degree required’ as it matches the search results page summary heading and is more concise.


Kateharries (approved):
Looks good

PR #732, opened 2022-10-26
Add post: Asking candidates to confirm reference requests

johndoates:
➡️ [Preview](https://deploy-preview-732--bat-design-history.netlify.app/apply-for-teacher-training/asking-candidates-to-confirm-who-reference-requests-should-be-sent-to/)


app/posts/apply-for-teacher-training/2022-10-24-asking-candidates-to-confirm-who-reference-requests-should-be-sent-to.md
```diff
@@ -0,0 +1,109 @@
...
+date: 2022-10-24
+---
+
+{% from "email/macro.njk" import appEmail %}
+
+We’ve [changed the references process](/changing-the-reference-process-to-make-it-easier-for-candidates-to-submit-applications) so that candidates:
+
+- only need to give details of people who can give references, rather than having to request and receive references before submitting their application
+- must confirm these details when they accept an offer, so that reference requests can be sent
+- can check on the progress of reference requests - this includes requesting more, cancelling requests and sending reminders
```
frankieroberto: I'm not sure this intro works. Could we make it more specific to this entry perhaps, and find other ways to link to the other entries?

Maybe something like this?

```suggestion
We’ve [changed the references process](/changing-the-reference-process-to-make-it-easier-for-candidates-to-submit-applications) so that references aren’t collected until after an offer is accepted. Candidates still have to [provide details of 2 people who can give a reference](/apply-for-teacher-training/asking-candidates-to-confirm-who-reference-requests-should-be-sent-to/) in the initial application form. When it comes to accepting an offer, they are given the chance to change these details.
```

app/posts/apply-for-teacher-training/2022-10-24-asking-candidates-to-confirm-who-reference-requests-should-be-sent-to.md
```diff
@@ -0,0 +1,109 @@
...
+
+## The issue
+
+We needed to design a way for candidates to confirm details of references when they accept an offer.
+
```
frankieroberto: Maybe we could drop this line? I feel like it's covered by the next para?

```suggestion
```

app/posts/apply-for-teacher-training/2022-10-24-asking-candidates-to-confirm-who-reference-requests-should-be-sent-to.md
```diff
@@ -0,0 +1,109 @@
...
+The service sends an email to a candidate when:
+
+- they accept a conditional offer
+- they accept an unconditional offer
+
+When the candidate accepts their offer, the service sends an email requesting references to each of the people whose details they gave.
+
+## Further work
+
+We need to add a message about references for unconditional offers, to match the one we show for conditional offers.
```
frankieroberto: Should we add something here about making it a bit clearer on the offer page that there’s the "Accept offer and conditions" radio option doesn’t do this straight away, but instead leads to the new confirm references page?

frankieroberto (commented):
Have added screenshots (which still need alt text). Looking good.

app/posts/apply-for-teacher-training/2022-10-24-asking-candidates-to-confirm-who-reference-requests-should-be-sent-to.md
johndoates: Yes, I'll add something about that.

frankieroberto (approved):
I've updated the screenshot, as the reference cards are now full-width. 

I think this is good to publish now?

PR #737, opened 2022-10-28
Simplify breadcrumbs: don’t show current page and use 'Home' for homepage

frankieroberto:
This updates the breadcrumb navigation to remove the current duplication of the name of the home page (which currently repeats the title in the black bar) and the name of the current page (which is also shown in the h1).

This follows the [GOV.UK Design System guidance for breadcrumbs](https://design-system.service.gov.uk/components/breadcrumbs/), which says:

> The breadcrumb should start with your ‘home’ page and end with the parent section of the current page.

(This change was also made [upstream in the GOV.UK Design History template repository](https://github.com/x-govuk/govuk-design-history/pull/188).)

### Service page

| Before | After |
|---|---|
|  localhost_8080_apply-for-teacher-training_(iPhone 6_7_8) (2) | localhost_8080_apply-for-teacher-training_(iPhone 6_7_8) (1) |

### Post page

| Before | After |
|---|---|
| localhost_8080_apply-for-teacher-training_letting-candidates-deal-with-reference-requests-after-accepting-an-offer_(iPhone 6_7_8) (1) | localhost_8080_apply-for-teacher-training_letting-candidates-deal-with-reference-requests-after-accepting-an-offer_(iPhone 6_7_8) |

### DIFF
--- a/app/_components/breadcrumbs/_breadcrumbs.scss
+++ b/app/_components/breadcrumbs/_breadcrumbs.scss
@@ -16,11 +16,6 @@ $app-breadcrumbs-inverted-border-colour: govuk-colour("light-blue");
     border-color: $app-breadcrumbs-inverted-text-colour;
   }
 
-  .govuk-breadcrumbs__list-item:last-child a {
-    font-weight: bold;
-    text-decoration: none;
-  }
-
   a:link,
   a:visited,
   a:hover,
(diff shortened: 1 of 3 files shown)


edwardhorsford (approved):

simonwhatley:
Can you fix the breadcrumb, so it doesn’t show two services if the post slug is the same?

For example:

- /find-teacher-training/**updating-information-about-bursaries-and-scholarships**/
- /publish-teacher-training-courses/**updating-information-about-bursaries-and-scholarships**/

It doesn’t seem to recognise that the parent is different.

image

simonwhatley (approved):

PR #761, opened 2022-11-11
Show images in index pages

frankieroberto:
This adds images to the index pages, where they are specified (as an `ogImage`) on the post.

Main reason is to add a bit of visual interest and possibly to make it easier to find relevant posts. Downside is it makes the page longer.  Possibly the image aspect ratios should a bit more wide (currently 3:2).

## Screenshot

design-history-with-images


frankieroberto:
Closing this as the team weren't keen.

PR #762, opened 2022-11-11
Add post: Adding a submit button to the ‘sort by’ form to improve accessibility

simonwhatley:


frankieroberto:
@simonwhatley interesting! Are you also making the filters and the sort part of the same `<form>`, so that you don't lose any options if changing them both and then hitting one of the green buttons? Or is this just unlikely to be an issue in practice?

simonwhatley:
@frankieroberto in the prototype and live, they're two forms. We maintain the state of all form fields between the sort and filter options, so if one is submitted, the other is not reset.

PR #787, opened 2022-12-15
Adding support for tags on posts

frankieroberto:
Adding support for "tags" on posts, and tagging all the existing posts for Apply.

➡️ [Preview](https://deploy-preview-787--bat-design-history.netlify.app/tags)

### Post page showing tags in header

Screenshot 2022-12-16 at 15 03 24

### Tag page

Screenshot 2022-12-16 at 15 03 15

### Page listing all tags

Screenshot 2022-12-16 at 15 03 02


lib/filters/post-tags.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Filter the tags on a post to remove the default ones
+ * we just use for creating the collections
+ * @param {Array} array Array to filter
+ * @return {Array} filtered array
+ */
+module.exports = (array) => {
+  const tagsToIgnore = ['post', 'apply-for-teacher-training', 'manage-teacher-training-applications', 'register-trainee-teachers', 'find-teacher-training']
```
simonwhatley: This tag list differs from the `ignoredTags` in `lib/filters/all-tags.js`. Should they be the same?

lib/filters/post-tags.js
```diff
@@ -0,0 +1,10 @@
+/**
+ * Filter the tags on a post to remove the default ones
+ * we just use for creating the collections
+ * @param {Array} array Array to filter
+ * @return {Array} filtered array
+ */
+module.exports = (array) => {
+  const tagsToIgnore = ["post", "apply-for-teacher-training", "manage-teacher-training-applications", "register-trainee-teachers", "find-teacher-training"]
```
edwardhorsford: Should this array be dynamically generated from somewhere?

app/tag.md
```diff
@@ -0,0 +1,9 @@
+---
+layout: tagged
+pagination:
+  data: collections
+  size: 1
+  alias: tag
+  filter: ['all', 'nav', 'blog', 'work', 'featuredWork', 'people', 'rss']
```
edwardhorsford: Feels like this list of tags to filter should / could combine with the file in `post-tags.js`?

app/_layouts/post.njk
```diff
@@ -11,7 +11,8 @@
       title: title | widont,
       date: page.date,
       modified: modified if modified,
-      tag: status
+      tag: status,
+      tags: (tags | postTags)
```
edwardhorsford: Might be tempted to rename this to be more about what the filter is doing.

MaeveRoseveare (approved):

lib/filters/post-tags.js
frankieroberto: @simonwhatley I've de-duplicated these now, moving the ignored tags to a JSON config file. Thanks to @edwardhorsford  for helping!

.eleventy.js
```diff
@@ -42,40 +42,78 @@ module.exports = function (eleventyConfig) {
...
-    })
+    return collection.getFilteredByGlob('app/posts/support-for-publish/*.md')
+  })
+
+  // A collection of reference pages
+  eleventyConfig.addCollection('reference', collectionApi => {
+    return collectionApi.getFilteredByGlob(['app/glossary.md', 'app/mission-patches.md', 'app/service-map.md'])
+  })
+
+  // A collection of reference pages
```
edwardhorsford: This comment needs updating

.eleventy.js
```diff
@@ -42,40 +42,78 @@ module.exports = function (eleventyConfig) {
...
+    // Set up empty list of tags
+    let tags = []
+
+    for (post of posts) {
+      if ('tags' in post.data) {
+        // Add any new tags from the post to the array
+        for (tag of post.data.tags) {
+          if (!tags.includes(tag)) {
+            tags.push(tag)
+          }
```
edwardhorsford: I think this might be case sensitive?

app/tags.html.njk
```diff
@@ -0,0 +1,12 @@
+---
+layout: page
+eleventyExcludeFromCollections: true
+permalink: "/tags/index.html"
+title: Tags
+---
+
+<ul class="govuk-list govuk-list--bullet">
+  {% for tag in collections.allTags %}
+    <li><a href="/tagged/{{ tag | slugify }}">{{ tag }}</a> ({{ collections[tag] | length }})</li>
```
edwardhorsford: Should this be using the slug filter?

app/_components/article/template.njk
```diff
@@ -24,8 +24,18 @@
             <span aria-hidden="true">•</span>
             Last updated <time datetime="{{ params.modified | date }}">{{ params.modified | date("d LLLL y") }}</time>
           {% endif %}
+
+          {% if params.header.tags | length > 0 %}
+            <span aria-hidden="true">•</span>
+            Tags:
+            {% for tag in params.header.tags %}
+              <a href="/tagged/{{ tag | slugify }}" class="govuk-link">{{ tag }}</a>
```
edwardhorsford: Should this be using the `slug` filter?

edwardhorsford (changes requested):

.eleventy.js
frankieroberto: Fixed, thanks.

.eleventy.js
frankieroberto: Yeah it is. I think I’m ok with that for now. Lowercasing everything would mean "UCAS" becomes "ucas", and doing a case-insensitive check when building the array would mean that whichever version appeared first would be how it appears in the tags list.  This way would at least let us spot when we’ve been consistent with the casing...?

edwardhorsford: It'll mean we have case sensitive urls, and would try to generate pages at `/tags/Foo` and `/tags/foo`.

frankieroberto: ok so I’ve added some extra code which will throw a helpful error if this happens so that we can catch it.

edwardhorsford (approved):
Great work @frankieroberto!

PR #811, opened 2023-01-26
Add new post for bulk awarding of teaching status

edwardhorsford:


edwardhorsford:
[Design history for bulk qts](https://trello.com/c/S0jhSKOX/5125-design-history-for-bulk-qts)

PR #823, opened 2023-02-07
Add post: Adding multiple locations to an organisation

simonwhatley:


simonwhatley:
[Adding locations in Support](https://trello.com/c/mN8LztJ0/1007-adding-locations-in-support)

simonwhatley:
@Kateharries I wonder whether I separate the editing locations (view, edit and remove) part of the design history into a separate post. What do you think?

Kateharries:
> Yes, it would make sense to put that in a separate design history so this doesn't become too long

Kateharries (approved):
Looks good

PR #824, opened 2023-02-09
Add post about editing HESA records

edwardhorsford:


edwardhorsford:
[Write design history entry for editing HESA records](https://trello.com/c/TqNOQ2F9/5246-write-design-history-entry-for-editing-hesa-records)

PR #870, opened 2023-03-29
Post for highlighting problems with records

edwardhorsford:


edwardhorsford:
[Write design history for showing problems on records](https://trello.com/c/ZC8U9APz/5338-write-design-history-for-showing-problems-on-records)

PR #874, opened 2023-04-06
Add post about allowing providers to import applications from Manage with conditions pending

johndoates:


app/posts/register-trainee-teachers/2023-04-06-allowing-providers-to-import-applications-with-conditions-pending.md
```diff
@@ -0,0 +1,96 @@
...
+---
+
+The integration between the Manage and Register services automatically imports applications from Manage when they reach recruited status. This happens when the candidate has either:
+
+- accepted an unconditional offer
+- met all the conditions of their offer
+
+We use recruited status to decide which applications to import because it’s the best way to identify someone who is due to start training.
+
+The integration is not active throughout the year. It’s turned on part way through each recruitment cycle, to help providers register trainees who will start training in the upcoming academic year.
```
edwardhorsford: This isn't right. It *is* on throughout the year, but won't start taking new cycle's applications till we tell it to. Older cycle ones still import.

app/posts/register-trainee-teachers/2023-04-06-allowing-providers-to-import-applications-with-conditions-pending.md
```diff
@@ -0,0 +1,96 @@
...
+
+The integration was used for some school centred initial teacher training (SCITT) applications in the 2020 to 2021 recruitment cycle and for the majority of them in 2021 to 2022.
+
+## Why we’re doing this
```
edwardhorsford: This heading refers to 'this', but we're no longer introducing what 'this' is - we need to open by saying something like 'we're expanding the applications that import' or something.

johndoates: Good point. I moved some content around which was previously in the introduction. Rather than adding context, I think I'll make the heading more specific.

app/posts/register-trainee-teachers/2023-04-06-allowing-providers-to-import-applications-with-conditions-pending.md
johndoates: I can rewrite it to be more accurate but do users actually see a difference? Won't applications from the previous cycle already have been imported during the previous cycle?

edwardhorsford (approved):

PR #883, opened 2023-04-18
Add post: Removing school placement vacancies in Find

simonwhatley:


simonwhatley:
[Remove school placement vacancies in Find](https://trello.com/c/HoKCkoIp/1237-remove-school-placement-vacancies-in-find)

Kateharries (approved):
Looks good

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
```diff
@@ -0,0 +1,66 @@
...
+
+We show all ‘Open’ and ‘Closed’ courses if the filter is deselected.
+
+![Applications open filter](applications-open-filter.png "Applications open filter")
+
+## Further considerations
+
+If someone deselects the ‘Applications open’ filter, we show all open and closed courses. It is not easy to see which courses accept applications and which do not.
+
+In the future, we could add a label to ‘Closed’ courses on the courses list so that users can quickly see whether they can apply for the course.
```
defong: Do we need to consider `Date you can apply from` 

image

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
```diff
@@ -0,0 +1,66 @@
...
+When a provider publishes a course, they can list all the schools where a candidate can gain experience. Previously, this list also showed whether there was a vacancy at the school.
+
+The vacancy status was a simple ‘yes’ or ‘no’ and required the provider to keep the course current.
+
+We removed the vacancy status because:
+
+- candidates do not apply to a vacancy at a school
+- providers often do not know the number of vacancies at their partner schools until late in the recruitment cycle
+- the number of vacancies can change during the recruitment cycle because candidates may need to withdraw
+- we are providing false hope that candidates will be placed at certain schools
```
defong, line 23: I don't think this has much bearing, as it still in an ambiguous  state

We still showing a list of schools with or without the potential of being placed in the desired school

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
```diff
@@ -0,0 +1,66 @@
...
+
+- school placements section of the course details
+- warning text when a course is closed
+- vacancies filter on the course list
+
+### School placements section
+
+We replaced the ‘Schools’ table with a list of schools and removed each school’s vacancy status.
+
+![School placements](school-placements.png "School placements")
```
defong, line 40: Should it still say 

> You can select one of the following schools to indicate a preferred placement when you apply.

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
```diff
@@ -0,0 +1,66 @@
...
+
+![Course closed for applications](warning-message.png "Course closed for applications")
+
+### Vacancies filter
+
+If a course is open, it is available on Find, and candidates can apply.
+
+We replaced the ‘Vacancies’ filter with one labelled ‘Applications open’. The filter is selected by default, so we ‘Only show courses accepting applications.’
+
+We show all ‘Open’ and ‘Closed’ courses if the filter is deselected.
```
defong: Maybe the use of `Open` and `Closed`  is too concise and not succinct given

> Only show courses accepting applications

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
simonwhatley: The messaging above the Schools list falls into a separate work that @Kateharries is doing.

The text could say:

> We work with the following schools to provide your school placements.

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
simonwhatley: @defong can you elaborate on what we need to consider? For example, do you think we need to change the information if the course is closed? 

Since `Date you can apply from` corresponds to the `Applications open date` in Publish it is possible candidates might be confused if the course is now closed. /cc @Kateharries

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
simonwhatley: @defong These are the reasons garnered from research regarding why vacancies are problematic. Can you elaborate on what is an ambiguous state and how we might mitigate against the ambiguity?

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
simonwhatley: @defong do you mean there is a disconnect between the `Open` and `Closed` states and the `Applications open` filter, which has the single option `Only show courses accepting applications`? What would be a better filter?

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
defong: It still ambiguous as the list of school shown has no bearing to being placed in.

I think we need to reword/justify the `false hope` symptom.

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
defong: Seems related to
https://github.com/DFE-Digital/bat-design-history/pull/883#discussion_r1180779642

So if the course can not be applied to is the course still `Only show courses accepting applications`

victoriawinlow (commented):
@simonwhatley I think the 'Applications open' filter will confuse users (both candidates and providers) with the 'Date you can apply from' field on the course page. I think the current 'Vacancies' filter is fine to leave as-is as we're looking at courses with vacancies here (rather than school placement vacancies), so that's still valid.

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
Kateharries: Yes this is a good point. Can we hide 'Date you can apply from' for closed courses? Ideally, we'd test this with candidates to see if they'd prefer to have that info there (even if the course is closed) or if it's just confusing to have it there. I think we should hide it if the course is closed.

Kateharries:
> @simonwhatley I think the 'Applications open' filter will confuse users (both candidates and providers) with the 'Date you can apply from' field on the course page. I think the current 'Vacancies' filter is fine to leave as-is as we're looking at courses with vacancies here (rather than school placement vacancies), so that's still valid.

I agree that we should hide 'Date you can apply from' if the course is closed. I prefer the new wording for the filter though, as it's more direct (you can either apply or you can't). We're removing the concept of courses having 'vacancies' so it removes any idea that there are X number of spaces. You can either apply or you can't.

Kateharries (approved):
As discussed, we'll add: - closed for applications

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
Kateharries: We're going to leave the date where it is and add: '- closed for applications' next to the date for courses that are closed.

app/posts/find-teacher-training/2023-04-18-removing-school-placement-vacancies.md
simonwhatley: > So if the course can not be applied to is the course still `Only show courses accepting applications`

If the course cannot be applied to, it will not show in the courses list if the user has checked `Only show courses accepting applications` in the filters. Only courses open for applications are shown in the list.

This is the same interaction as the old vacancies filter where checking `Only show courses with vacancies` only shows courses open for applications in the list.

If the user knows the course URL and goes directly to a closed course, or has unchecked the filter, the apply button is replaced by the closed for applications message.

simonwhatley:
For closed courses, next to the 'Date you can apply from' date we have appended '- closed for applications':

image

PR #934, opened 2023-06-05
Remove reference to old BAT design history URL

simonwhatley:
This PR removes `https://bat-design-history.netlify.app` from URLs.


frankieroberto (approved):
Thanks!

PR #950, opened 2023-06-14
Add post: Improving how users add schools

simonwhatley:


simonwhatley:
[Improving how schools are added in Support](https://trello.com/c/9VVyJ1Dw/1345-improving-how-schools-are-added-in-support)

PR #955, opened 2023-06-15
Add twitter cards markup

simonwhatley:
See details of markup: https://developer.twitter.com/en/docs/twitter-for-websites/cards/overview/markup


frankieroberto (approved):
Thanks!

PR #986, opened 2023-08-07
Publish: Added post about improving the 'withdraw a course' page

MaeveRoseveare:
Added a post to Publish based on Kate's work to improve the withdraw page. Kate drafted the post, I just added it.


MaeveRoseveare:
[Design history: improving content on the 'Withdraw course' page.](https://trello.com/c/vzUPFHdY/427-design-history-improving-content-on-the-withdraw-course-page)

MylesJarvis (approved):
LGTM! I do wonder whether there's a way we could make it more difficult to withdraw from where we put the button on the course page. Or even (I know this sounds like heresy) but adding even more friction to this journey through another window to _really_ get them to stop and think.

MaeveRoseveare:
> LGTM! I do wonder whether there's a way we could make it more difficult to withdraw from where we put the button on the course page. Or even (I know this sounds like heresy) but adding even more friction to this journey through another window to _really_ get them to stop and think.

yeah something to think about. The consequences seem quite big for candidates if they've applied. So one to watch to see if there's more we can do.

PR #1015, opened 2023-10-12
Add post: Problems with school placements

simonwhatley:


simonwhatley:
[Design History - Problems we with school placements](https://trello.com/c/TQJi16b7/28-design-history-problems-we-with-school-placements)

PR #1030, opened 2023-10-23
Apply: post about better content in between cycles

MaeveRoseveare:
Post about showing candidates clearer content in between recruitment cycles.

https://trello.com/c/SG1N5OF4/815-apply-content-we-show-in-between-each-cycle-for-continuous-applications


MaeveRoseveare:
[Design history: Apply: content we show in between each cycle for continuous applications](https://trello.com/c/SG1N5OF4/815-apply-content-we-show-in-between-each-cycle-for-continuous-applications)

Kateharries (approved):
looks good

PR #1045, opened 2023-10-31
Update post: Reordering the filter list

MaeveRoseveare:
Edited the section about the order of the filters matching the order of the list item for accuracy.


simonwhatley:
@MaeveRoseveare I think the change is less about logic and more about needs

So from this:

> we believed we could order them in a more logical way based on the high number of international candidates we get.

to something like this:

> we believed we could order them in a way that better matched the needs of the high number of international candidates we get.

simonwhatley (approved):
It _feels_ like we should handle international candidate needs better. If we must prevent them from applying to courses that don't sponsor student visas, this should be addressed in Apply.

PR #1049, opened 2023-11-01
Apply: helping candidates understand why they cannot submit their applications

MaeveRoseveare:
Added post about content we created when a candidate cannot submit an application.

https://trello.com/c/z68XCOkA/901-apply-content-we-show-when-a-candidate-cannot-submit-an-application


lsalivonaite-transform (approved):
It's perfect!

PR #1097, opened 2024-01-23
Apply: Enhancing the UX and UI for application pages

lsalivonaite-transform:
New design history for Apply service outlining the changes we have done to 'Your applications' page and different status pages, with the introduction of new pattern 'what happens next'


kelliedesigner (approved):
Looks good to me

PR #1101, opened 2024-01-30
Add post: Why are we working on ‘Claim funding for mentors’?

simonwhatley:


simonwhatley:
[Design histories - Discovery and alpha](https://trello.com/c/nZ4eEVt7/99-design-histories-discovery-and-alpha)

PR #1127, opened 2024-02-28
Add post: Why we are no longer reliant on placement data from Register

simonwhatley:


simonwhatley:
We will use data from 'Register trainee teachers' to prepopulate the service.

PR #1139, opened 2024-03-21
Add post: Round 1 research findings

simonwhatley:


olioldfield:
Think the apostrophes need to be moved after the s:

Explore user’s expectations of how they expect to be notified of the new service
Explore user’s expectations of the guidance and the level of support they will require with this being a brand-new service

olioldfield:
Would it be possible to remove the repetition of as a school:

As a school
I need to know what the timelines for claiming funding are
So that I know when we will receive our payment

As a school
I need reassurance that the DfE have the correct details for our organisation
So that I feel confident that our claim has been submitted correctly

As a school
I need to know what will happen after my claim has been submitted
So that I feel informed of the process and know what to expect from the next steps

As a school
I need to understand how to use the service and know what is required of me before I start a claim
So that I know what to expect and can prepare all the information required information ready before I start our claim

As a school
I need to know the status of my claim after it has been submitted
So that I can track the progress of our claim

As a school
I need to know if there is an issue with my claim
So that I can rectify it

olioldfield (approved):

simonwhatley:
[Round 1 user research - findings write up](https://trello.com/c/LoeqaEPM/306-round-1-user-research-findings-write-up)

PR #1148, opened 2024-03-27
Add post: Logging who submitted a claim

simonwhatley:


simonwhatley:
[As part of assurance it needs to be clear which user submitted a claim](https://trello.com/c/M7EhoxFO/263-as-part-of-assurance-it-needs-to-be-clear-which-user-submitted-a-claim)

olioldfield (approved):

PR #1191, opened 2024-05-17
Add post: User research with private beta ITT providers

simonwhatley:


olioldfield (commented):
Change: 
Key insights 
Overarching need for providers 
Providers need to be able to view the claims submitted by schools and validate them against their records.

To:

Key insights 
Overarching need for providers 
Providers need to be able to view the claims submitted by schools and validate them against their records if they wish.
The process for doing this needs to be simple and not add too much administrative burden.
Providers also need clarity on the process to provide evidence when a claim has been selected for post-payment auditing.

olioldfield (approved):

PR #1197, opened 2024-05-24
Add post: How we onboarded schools in private beta

simonwhatley:


olioldfield (commented):
Looks great - just one small change required:

Current wording:

Both providers need to gain consent from the user first to provide DfE with a contact and email address.

Updated to

Both providers needed to gain consent from the user first to provide DfE with a contact and email address.

olioldfield (approved):

PR #1200, opened 2024-05-28
Add post: Emailing support users

simonwhatley:
We email support users when they have been added or removed from the service by another support user.


olioldfield (approved):
Very nice - love the reference to Slack :)

PR #1213, opened 2024-06-03
Replace the node-sass package with sass package

simonwhatley:
Node Sass is now deprecated. While it will continue to receive maintenance releases indefinitely, there are no plans to add additional features or make it compatible with any new CSS or Sass features. Projects that still use it should move onto Dart Sass.


andrewscrivener:
Thanks @simonwhatley , this is great. I've pulled the repo locally to test and all looks good to me.

There are 34 Dart Sass deprecation warnings for the govuk-frontend module that we'll need to consider for future updates. such as the following;

```Deprecation Warning: Using / for division outside of calc() is deprecated and will be removed in Dart Sass 2.0.0.```

:shipit:

andrewscrivener (approved):

PR #1214, opened 2024-06-03
Move JavaScripts to the assets folder

simonwhatley:
We have `fonts`, `images` and `styles` in the `/assets` folder. To be consistent, this PR moves the JavaScripts there too.


andrewscrivener (approved):
Yes, I agree with moving them for consistency. 👍

PR #1219, opened 2024-06-10
Embed user needs into posts

simonwhatley:
The user needs section for Apply was removed in a long-ago update to the design history site. This resulted in the remaining user needs functionality being broken.

This PR moves the user needs into each post, rather than returning the design history site to its previous state. This is because:

- a long time has passed since the breaking change was done, so it is easier to embed the user needs rather than restore functionality
- embedding the user needs simplifies each post - there is no need for extra metadata or dependencies
- in future, we hope to combine directories for images and posts, and this work lays the foundation for a simpler post architecture

In the future, we may consider splitting out user needs again and making them function similarly to tags.


andrewscrivener (approved):
I've pulled the branch down locally to test and all looks good to me. Here are a couple of screengrabs of the user need component. :shipit: 

User need example 1

User need example 2

PR #1221, opened 2024-06-11
Reorganise Find and Apply post subdirectories

simonwhatley:
This PR moves posts in the Find and Apply subdirectories into the main directories for each service.

There is no need for the posts to be separate from the main post timeline in each section.

This change is part of an ongoing effort to reorganise and clean up the design history site so that we can merge post and image directories more easily.


andrewscrivener (approved):
I've pulled the branch and reviewed locally as well as looking at the diffs. All makes sense. :shipit:

PR #1278, opened 2024-07-30
Update headers to improve site security

simonwhatley:
Whilst it's optional to use these headers as the design history site is statically generated, it is good practice.

Mozilla Observatory gives the site an A+ score if we implement these headers:

https://observatory.mozilla.org/analyze/deploy-preview-1278--bat-design-history.netlify.app

### DIFF
--- a/netlify.toml
+++ b/netlify.toml
@@ -1,3 +1,18 @@
+[[headers]]
+  for = "/*"
+  [headers.values]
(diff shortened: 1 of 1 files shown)


simonwhatley:
@Nitemaeric, can you review the content security policy in the `netlify.toml` file and tell me what I need to do to set it up correctly? I've commented it out for now as it was too restrictive.

Nitemaeric (approved):

PR #1281, opened 2024-08-01
Add post: Quietly re-opening the service to new claim submissions in July and August

simonwhatley:


simonwhatley:
[Why we experimented with quietly re-opening the claim window at the end of July DH](https://trello.com/c/xSdwj1xV/615-why-we-experimented-with-quietly-re-opening-the-claim-window-at-the-end-of-july-dh)

olioldfield (approved):
Thank you!

PR #1288, opened 2024-08-19
Add TRA services to the service map

simonwhatley:
This PR adds two TRA services to the service map:

- Check a teacher’s record
- Check the children’s barred list


app/images/service-map/service-map.png
ralph-hawkins, line 1: Is it worth cropping the bottom pixels off, give they're just empty?

app/service-map.md
```diff
@@ -129,7 +129,15 @@ The ‘Becoming a teacher’ services are part of a broader range of digital ser
...
-        "name": "Employer Access",
+        "name": "Check a teacher’s record",
+        "href": "https://check-a-teachers-record.education.gov.uk"
+      },
+      {
+        "name": "Check the children’s barred list",
+        "href": "https://check-the-childrens-barred-list.education.gov.uk"
+      },
+      {
+        "name": "Employer Access - closing soon",
```
ralph-hawkins, line 140: Do we know when its closing?

ralph-hawkins (approved):
Looks good, a couple of minor comments. I wonder if this should live in general 'Become a teacher' section of the site?

app/service-map.md
simonwhatley: We only know it's closing soon.

app/images/service-map/service-map.png
simonwhatley: I've now trimmed some pixels as suggested

PR #1291, opened 2024-08-23
Add post: Defining the sampling algorithm

simonwhatley:


simonwhatley:
[Sampling algorithm](https://trello.com/c/CEe1vWBF/606-sampling-algorithm)

simonwhatley:
@olioldfield, do you need to add something to the post that highlights the sampling will be done outside the service in Private Beta, but in Public Beta it will be brought into the service?

olioldfield:
> @olioldfield, do you need to add something to the post that highlights the sampling will be done outside the service in Private Beta, but in Public Beta it will be brought into the service?

we could write, that the algorithm is currently run off service to produce a csv that is imported and for public beta we will look to add this functionality into the service to enable the claim statuses to be updated without a csv upload.

PR #1293, opened 2024-08-23
Add post: Exporting claims in Support

simonwhatley:


olioldfield (commented):
@Nitemaeric @simonwhatley I think as we now have multiple exports, we probably need to break this down into multiple posts. 
Maybe a DH for Payments, one for Sampling and one for Clawback and the associated processes and csv exports and imports. As the exports will have slightly different CSV headers. What do you think?

simonwhatley:
@olioldfield, this is a different (older) post from the newer work on managing claims through the payments, sampling, and clawback processes. I agree that the latter three should also have design histories.

olioldfield (approved):

PR #1363, opened 2024-10-28
Add post: Managing payments in Support

simonwhatley:
See post: https://deploy-preview-1363--bat-design-history.netlify.app/claim-funding-for-mentors/managing-payments-in-support/


olioldfield:
Current:

Since the ESFA will not have access to Claim funding for mentor training, we decided to share information using email and CSV files.

Proposed change:

Upon initiating the private beta we scheduled regular catch-ups with the ESFA to discuss the service and their requirements. They advised that in order to make payments they would require a CSV file and stipulated that they needed the following information:

- school_urn
- school_name
- school_local_authority
- school_establishment_type
- school_establishment_type_group
- claim_amount

They also advised that they were building an API but it would not be ready for either the private or public beta.

olioldfield:
I don't think we ever considered giving the ESFA access - it was never mentioned in our meetings and they definitely didn't request it:

Further considerations 
We considered some changes that we did not implement. These included:

Giving the ESFA access to the service

olioldfield:
Perhaps we could add:

Future considerations:

The public beta team could consider giving the ESFA access to the service if they requested it.

olioldfield (approved):

PR #1370, opened 2024-10-29
Add post: Updating links to destructive actions

simonwhatley:


olioldfield (approved):
I personally prefer the red links, and think these get a little lost in the UI and run the risk that destructive actions may be chosen by mistake.

PR #1391, opened 2024-11-21
Add redirect for missing trailing slash

simonwhatley:


simonwhatley:
These redirect rules do not appear to work.

PR #1427, opened 2025-01-10
Add post: Approval for sharing information securely with the Education and Skills Funding Agency (ESFA) and accredited providers

olioldfield:
…ed-providers.md

Created the design history to detail the approval of our approach to share data with the ESFA and accredited providers


olioldfield:
Grace has approved via Slack

PR #1480, opened 2025-02-24
Addendum added

andrewscrivener:
Addendum added to "Improving the withdrawals journey" post.


andrewscrivener:
@simonwhatley I didn't realise this was an addendum to an existing post. What do you think about the publication date? Should it remain the original published date or be updated to today?

simonwhatley:
@andrewscrivener In my experience, it is better to retain the old design history as a snapshot in time and create a new design history with the updated information rather than amending existing posts. This keeps the narrative clean. The latest design history does not need to be long.

tehseenudin (commented):
Agree that it makes sense to post as a new design history and reference the previous one

PR #1484, opened 2025-02-26
User research round 11 findings

petejobeskinandcarta:


olioldfield:
Hiya @petejobeskinandcarta I was wondering if I could fix this as I am trying to improve my DH expertise, but maybe this needs to be closed or something else?

petejobeskinandcarta:
@olioldfield I think it can be merged now, and your changes should go through? There's a lint thing that says the headings are not heirachical but I can't see any that aren't so think we can just merge.

olioldfield:
> @olioldfield I think it can be merged now, and your changes should go through? There's a lint thing that says the headings are not heirachical but I can't see any that aren't so think we can just merge.

Yeah, agreed - merge away!!

petejobeskinandcarta:
Thanks - merged successfully :)

PR #1509, opened 2025-03-13
Add post: Changing provider details

simonwhatley:
Post URL: https://deploy-preview-1509--bat-design-history.netlify.app/register-of-training-providers/changing-provider-details/


olioldfield (approved):
Love it

tehseenudin (approved):
Really like this - especially how SCITT is derived. An observation, does 'contact support' need to be expanded to state 'contact BAT Support'?

andrewscrivener (approved):
It reads well to me. No further comments :shipit:

simonwhatley:
@tehseenudin 

> does 'contact support' need to be expanded to state 'contact BAT Support'?

I don’t think we need to change the design history. However, there’s a piece of work we need to do to add support information to the service. When we do the work we’ll clarify it’s BAT support.

olioldfield:
> @tehseenudin
> 
> > does 'contact support' need to be expanded to state 'contact BAT Support'?
> 
> I don’t think we need to change the design history. However, there’s a piece of work we need to do to add support information to the service. When we do the work we’ll clarify it’s BAT support.

Interesting- we should probably check with Julie B about what the team is actually called as they support services beyond BAT

tomtrentham (approved):
This looks great. Just noting my comment from the other PR - it might be useful to have URNs for HEIs as well.

PR #1514, opened 2025-03-13
Include ITT mentor services in the project description

carlosmartinez:


app/_data/app.js
```diff
@@ -1,6 +1,6 @@
 module.exports = {
   productName: 'Becoming a teacher design history',
-  description: 'A history of the designs for the Find, Apply, Publish, Manage, Register and Support services',
+  description: 'A history of the designs for the Find, Apply, Publish, Manage, Register, Manage school placements, Claim funding for mentor training and Support services',
```
simonwhatley: We should use an abbreviated version rather than the whole service name for 'Manage school placements' and 'Claim funding for mentor training.' This will follow the style used for the other services. For example:

```suggestion
  description: 'A history of the designs for the Find, Apply, Publish, Manage, Register, Placements, Claim and Support services',
```

app/index.md
```diff
@@ -1,7 +1,7 @@
 ---
 layout: collection
 title: Becoming a teacher design history
-description: A history of the designs for the Find, Apply, Publish, Manage, Register and Support services
+description: A history of the designs for the Find, Apply, Publish, Manage, Register, Manage school placements, Claim funding for mentor training and Support services
```
simonwhatley: As per my first comment. We should use an abbreviated version rather than the whole service name for 'Manage school placements' and 'Claim funding for mentor training.' This will follow the style used for the other services. For example:

```suggestion
description: A history of the designs for the Find, Apply, Publish, Manage, Register, Placements, Claim and Support services
```

package.json
```diff
@@ -5,7 +5,7 @@
   "engines": {
     "node": "20.11.0"
   },
-  "description": "A history of the designs for the Find, Apply, Publish, Manage, Register and Support services",
+  "description": "A history of the designs for the Find, Apply, Publish, Manage, Register, Manage school placements, Claim funding for mentor training and Support services",
```
simonwhatley: As per my first comment. We should use an abbreviated version rather than the whole service name for 'Manage school placements' and 'Claim funding for mentor training.' This will follow the style used for the other services. For example:

```suggestion
  "description": "A history of the designs for the Find, Apply, Publish, Manage, Register, Placements, Claim and Support services",
```

simonwhatley (changes requested):
For 'Manage school placements' and 'Claim funding for mentor training, ' we should use an abbreviated version rather than the whole service name. This abbreviation will follow the style used for the other services.

In future, we should consider replacing the list of services in the description with something that scales better.

olioldfield (commented):
I think the issue with abbreviations is that so many services share similar names, which is why we shortened claim funding for mentor training to funding mentors. But again, that isn't clear as really it needs to be funding ITT mentors. 

Probably need to do a bit of thinking around the abbreviations if we go down that route

app/_data/app.js
carlosmartinez: How about `A history of the designs for the Becoming a Teacher service line`?

olioldfield: I like this idea - very neat and allows for future expansion

simonwhatley: I think that tagline simply repeats the site's title, 'Becoming a teacher design history'.

It's worth expanding the title and scope from just ‘design’ to include research, product and development, etc.

simonwhatley: Something like this might work:

> Capturing the history of product, design, and technology decisions shaping trainee teacher recruitment—from finding courses and applying, to managing applications, placements, funding mentors, and registering trainee teachers.

'Capturing' could be replaced with 'Documenting' or something similar.

olioldfield: A history of the product, design, technology and policy decisions shaping the services that enable and support trainee teacher recruitment.

simonwhatley: Let's go with @olioldfield's suggestion.

simonwhatley (approved):

carlosmartinez:
thanks @olioldfield and @simonwhatley. merging.

PR #1516, opened 2025-03-14
Fix curly quotes

simonwhatley:
This PR attempts to replace straight quotes with curly (a.k.a. ‘smart’) quotes.

Not all quotes have been changed. Most double quotes still need to be carefully considered - they cannot be replaced in bulk since front matter and Nunjucks components use straight quotes.


defong (approved):
Approving before @tomtrentham wants payback

PR #1562, opened 2025-04-10
Add post: Why we will not reinstate deleted user accounts

simonwhatley:
Post URL: https://deploy-preview-1562--bat-design-history.netlify.app/register-of-training-providers/why-we-will-not-reinstate-deleted-user-accounts/


defong (approved):
Nice deliberation

olioldfield:
@simonwhatley what if we added a reason for deletion? Could be structured, e.g. user request, security concern, etc. Would that provide enough of a mitigation to enable deleted accounts to be reinstated? Combined with some rules, e.g. security concern accounts can't be reinstated. In addition, this would also provide a flag that a previously deleted account due to a security concern, was now trying to gain access again?

olioldfield (approved):

tomtrentham (approved):

PR #1564, opened 2025-04-11
Add post: Onboarding orgs that do not have a URN

olioldfield:
Deploy preview: https://deploy-preview-1564--bat-design-history.netlify.app/claim-funding-for-mentors/onboarding-organisations-that-do-not-have-a-unique-reference-number/


gracevdfe:
Thanks all - looks fine to me. Only tweak needed is that there's an extra space after 'Making the vendor number mandatory' title

simonwhatley:
> Thanks all - looks fine to me. Only tweak needed is that there's an extra space after 'Making the vendor number mandatory' title

@gracevdfe ¡Hola! The extra line after the heading is because each heading has an anchor link, which, depending on its length, creates what appears to be an additional line. The markdown is correct and doesn't include an extra line. This is something we probably need to fix in the code in some way, albeit it's _no biggie_.

olioldfield:
Thanks both - I have made that title longer, so the anchor link doesnt make it look like there is an extra space

PR #1579, opened 2025-04-24
Fix post tag functionality and associated pages

simonwhatley:
This PR:

- updates the allTags collection to take into account different cases for tags - for example, it combines `get into teaching` and `Get Into Teaching` into one tag
- generates tag counts so we don't need to use the Nunjucks filter
- adds a postsByTag collection to handle tags with different cases more effectively - for example, you can pass it `get into teaching` or `Get Into Teaching` and it will return all posts tagged with either
- adds tag layouts
- updates tag pages - for example, adds breadcrumbs to the Tags page


olioldfield (approved):
YESSSSS. Love this!

defong (approved):
Work as intended
image

PR #1588, opened 2025-04-28
Add post: Removing personal data from page titles

simonwhatley:
View post: [Removing personal data from page titles](https://deploy-preview-1588--bat-design-history.netlify.app/register-of-training-providers/removing-personal-data-from-page-titles/)


tehseenudin (approved):
Looks good - two minor comments:
1. PII v personal data - ODPO commented recently to me that in the UK we refer to personal data, not PII (which is more US based terminology)
2. How we approach it - 2nd bullet - would including hyperlinks to service manual, DPA and the good practice referenced be helpful?

olioldfield (approved):

PR #1604, opened 2025-05-14
Fix @11ty/eleventy at v3.0.0

simonwhatley:
This PR:

- fixes `@11ty/eleventy` at v3.0.0 and allows patch-level changes. v3.1.0 breaks the site. Without this change, Dependabot will automatically upgrade the package to the next minor version.
- downgrades `rollup` to v3 as v4 breaks the Netlify deploy! 😱


kelliedesigner (approved):
LGTM

PR #1610, opened 2025-05-16
Update service map with 'Record Inductions as an Appropriate Body'

simonwhatley:
This PR changes `Appropriate Body Portal` to `Record Inductions as an Appropriate Body`.

View the service map here: https://deploy-preview-1610--bat-design-history.netlify.app/service-map/


emily-prudence-dfe (approved):
Looks great, thank Simon!

PR #1621, opened 2025-05-23
Add post: Why we soft delete rather than hard delete

simonwhatley:
Post URL: [Why we soft delete rather than hard delete](https://deploy-preview-1621--bat-design-history.netlify.app/register-of-training-providers/why-we-soft-delete-rather-than-hard-delete/)


tomtrentham (approved):

tehseenudin (approved):
Looks good - another point to consider and we have on the roadmap under DPIA, will be data retention policies.

PR #1631, opened 2025-06-09
Add post: Naming the service - Register of training providers

simonwhatley:
Post URL: https://deploy-preview-1631--bat-design-history.netlify.app/register-of-training-providers/naming-the-service/

### DIFF
--- a/app/posts/register-of-training-providers/2025-01-30-naming-the-service.md
+++ b/app/posts/register-of-training-providers/2025-01-30-naming-the-service.md
@@ -1,0 +1,71 @@
+---
+title: "Naming the service: Register of training providers"
+description: Choosing the right name for a service is essential to its success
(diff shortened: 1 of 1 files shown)


olioldfield (approved):
Looks great - but it will be interesting to see if there is feedback about whether ITT or something similar needs to be added in at a later date

tehseenudin (approved):
Looks good - two comments:
1. the date under the title says 30 January 2025 - should it be different?
2. there appears to be an extra line underneath the last heading

simonwhatley:
@tehseenudin, in response to your questions:

1. I've backdated the post to when the service's name was decided
2. The additional line under the last heading is a result of the heading having an anchor link. If you hover the title, you'll see a `#` symbol which links to the heading anchor.

PR #1633, opened 2025-06-09
Add post: How we use heading captions

simonwhatley:
This post documents how we use heading captions in the Register of training providers.

Post URL: https://deploy-preview-1633--bat-design-history.netlify.app/register-of-training-providers/how-we-use-heading-captions/


defong (approved):

andrewscrivener (approved):
I agree with the points made and have no further comments.

PR #1634, opened 2025-06-10
Add post: How we show success, error and notification banners

simonwhatley:
This post documents how we show success, error and notification banners in the Register of training providers.

Post URL: https://deploy-preview-1634--bat-design-history.netlify.app/register-of-training-providers/how-we-show-success-error-and-notification-banners/


app/posts/register-of-training-providers/2025-06-10-how-we-show-success-error-and-notification-banners.md
```diff
@@ -0,0 +1,95 @@
...
+
+- error summaries
+- notification banners
+- success messages
+
+Each type plays a different role. We aim to use banners consistently and clearly to help users stay on track.
+
+## Error summaries
+
+We display an error summary at the top of the page, above the main heading, whenever a user submits a form with validation errors. We also show an inline error message next to each field that contains an error.
```
defong, line 29: I feel we should add 
> Error: ’ to the beginning of the page <title>

as it improves accessibility

app/posts/register-of-training-providers/2025-06-10-how-we-show-success-error-and-notification-banners.md
```diff
@@ -0,0 +1,95 @@
...
+
+## Notification banners
+
+We use notification banners to inform users of important information that is not directly related to the action they are currently taking.
+
+They appear at the top of the page, just above the main heading, and remain visible until we update or remove them. Users cannot dismiss notification banners.
+
+We use this pattern to:
+
+- warn about service-wide issues (such as maintenance or outages)
```
defong, line 51: I am unsure of that claim, we probably need a trello card to cover that

app/posts/register-of-training-providers/2025-06-10-how-we-show-success-error-and-notification-banners.md
simonwhatley: @defong, while your comment is valid, I didn't want to duplicate too much content from the design system in this post.

app/posts/register-of-training-providers/2025-06-10-how-we-show-success-error-and-notification-banners.md
simonwhatley: @defong, this is how we should use the pattern/component, even if we don't have a reason to at this point in time.

andrewscrivener (approved):
I understand you mention it at the end, but is it worth explicitly stating that although the banner is full-width, the content within it should always wrap at two-thirds for readability? Or do you mean that this is not the case in this service?

simonwhatley:
@andrewscrivener, you're correct. When using the notification banner full width, the content within it will show two-thirds of the width. The error summary does not behave the same way; when displaying the summary in full width, the content does not wrap at two-thirds of the width. This is a known issue and one we need to monitor closely. The service only has one error message, which could be a problem: "Enter a Department for Education email address in the correct format, like name@education.gov.uk". I'll update the post to be more explicit.

defong (approved):

PR #1648, opened 2025-06-19
New post: changes to the locations flow in find a candidate

RebeccaCryan:
New post detailing the content and interaction changes to the location preference setting flow in Find a candidate


RebeccaCryan:
@PeteWilliams I feel like I've put the images in exactly the same way as in the last but they are not rendering. Do you mind having a look?

PR #1674, opened 2025-07-09
Creating external linking design history for teacher success

EllieNodder:


jesshannington:
Looks great- just a few things:

- consistency with capitalisation of 'Department'- if we're doing GOV.UK style we can just have lower case when it isn't 'Department for Education'
- do we need to say where the disclaimer is going to appear?
- do we have a plan to review the links? If so, who is going to do it?

PR #1803, opened 2025-10-23
Fix broken links on design history post

edwardhorsford:
This code and post were written with the intention they could be used by other teams - alas it seems the links no longer worked - this updates them to the current location.


simonwhatley (approved):
Looks good to me 🙌

PR #1835, opened 2025-11-28
Add post: Making contact phone optional

simonwhatley:
View post: https://deploy-preview-1835--bat-design-history.netlify.app/register-of-training-providers/making-contact-phone-number-optional/


tehseenudin (approved):

defong (commented):
> we have not found a need

Why did we not chose to remove it instead of making it optional?

simonwhatley:
> > we have not found a need
> 
> Why did we not chose to remove it instead of making it optional?

While I state "we have not found a need", that's not entirely accurate. We know Publish has/had phone numbers, and we also know Market regs need contact details. By making the phone number optional, we retain the ability to collect the details.

I will update the post.

PR #1897, opened 2026-02-13
Add post: Seeding the register with provider data

simonwhatley:
View post: https://deploy-preview-1897--bat-design-history.netlify.app/register-of-training-providers/seeding-the-register-with-provider-data/


tehseenudin (approved):
This looks good! I think somewhere (either in this DH or a separate one) it would be valuable to talk about the PII side of data, and the work that we considered to process PII work securely but had to be descoped due to complexity for dfE estate

simonwhatley:
> This looks good! I think somewhere (either in this DH or a separate one) it would be valuable to talk about the PII side of data, and the work that we considered to process PII work securely but had to be descoped due to complexity for dfE estate

Since we haven't imported contact data (and therefore PII), I'll discuss it in the related post later.

defong (commented):
Looks pretty correct to me

